PR #8, opened 2021-09-07
feat(desktop-entries): Deduplicate results from different sources

mmstick:
Prefer entries from local sources before system sources.

Can replicate by installing GIMP from apt and flatpak, and seeing if the flatpak is the only choice in the launcher.

### DIFF
--- a/plugins/src/desktop_entries/mod.rs
+++ b/plugins/src/desktop_entries/mod.rs
@@ -26,13 +26,12 @@ struct Item {
 impl Hash for Item {
     fn hash<H: Hasher>(&self, state: &mut H) {
         self.name.hash(state);
-        self.src.hash(state);
     }
 }
 
 impl PartialEq for Item {
     fn eq(&self, other: &Self) -> bool {
-        self.name == other.name && self.src == other.src
+        self.name == other.name
     }
 }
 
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

jacobgkau (changes requested):
Before the change, Flatpak and system sources were both displayed and clearly marked:

Screenshot from 2021-09-07 09-11-01

After the change, only the Flatpak version is displayed:

Screenshot from 2021-09-07 09-11-36

Meanwhile, the GNOME Shell Applications menu continues to show both versions, but without differentiating them:

Screenshot from 2021-09-07 09-11-16

So this seems to be removing functionality. What is the rationale for this change? If we think Flatpaks should take priority, can we just change the sort order without hiding the system version? (If a user doesn't ever want to see one version, they can just uninstall it.)

mmstick:
Requested behavior from https://github.com/pop-os/shell/issues/1006#issuecomment-913967891
Although I'm fine with not having this merged.

jacobgkau:
Looking at that issue, they're not using a Flatpak as an example, they have a system app and a "user" app that's most likely a .desktop file they copied to `~/.local/share/applications/` for some reason. (I can see people copying .desktop files to `~/.local/share/applications` to e.g. change an app's startup parameters.)

It seems like sorting local before system would be reasonable, but if we want to try and hide duplicates, I think filtering on .desktop filename would be more reasonable than however it's currently working (looks like by comment right now?) Filtering by .desktop filename would allow someone to copy the .desktop file, modify the exec line, and only have their new version show up (but would still display `org.gimp.GIMP.desktop` and `gimp.desktop` separately.)

mmstick:
The current behavior is filtering by the name and source, so you can't have a duplicate in the same source, but you can have duplicates across sources. This change still filters by name, but removes the source distinction. I'm not sure if there's a good way to determine if a file is a modified variant that should replace another.

jacobgkau:
According to the FreeDesktop specification for desktop files (linked to in [this comment](https://github.com/pop-os/shell/issues/1163#issuecomment-914678287) by another commenter): https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#desktop-file-id 

```
To determine the ID of a desktop file, make its full path relative to the $XDG_DATA_DIRS component in which the desktop file is installed, remove the "applications/" prefix, and turn '/' into '-'.

For example /usr/share/applications/foo/bar.desktop has the desktop file ID foo-bar.desktop.

If multiple files have the same desktop file ID, the first one in the $XDG_DATA_DIRS precedence order is used.
```

This seems to confirm that .desktop filename (or a derivative thereof) is what should be used to deduplicate, which I believe is what the GNOME Applications menu and favorites bar do. Just posting here because that seems equally relevant to this PR compared to the issue it was posted in.

PR #14, opened 2021-09-08
fix(service): Stale search results after interrupt

### LINKED ISSUE #13: Calculator, find plugins don't update results if characters are typed too quickly
I had this weird issue today.
I opened the launcher with the pop os shortcut, then proceeded to type =109-30. The result I got was 106. I tried it with a few different numbers, for example with a 40 it also gave me the wrong result. After tinkering around a bit, including using the normal numbers (not on the numblock), it worked for the 30 again, regardless if I typed the 30 on the numblock or not. 
It seems to onluy happen with numbers between 100 and 110(at least when I tested I could only reproduce it there, but I'm not sure if that's one condition). 

OS: Arch
Package source: `aur/gnome-shell-extension-pop-shell-git`
Are there any other important information I can supply?

mmstick:
Closes #13

### DIFF
--- a/service/src/lib.rs
+++ b/service/src/lib.rs
@@ -322,6 +322,7 @@ impl<O: futures::Sink<Response> + Unpin> Service<O> {
             if !self.search_scheduled {
                 self.interrupt().await;
                 self.search_scheduled = true;
+                self.last_query = query;
             }
 
             return;


jacobgkau (approved):
Fixes https://github.com/pop-os/launcher/issues/13. Regression testing passed on Hirsute and Focal.

PR #18, opened 2021-09-13
fix(calc): Update description from JS to Rust dependency

friday:
You might want it to say something different though, but at least not Math.JS any longer :)

### DIFF
--- a/plugins/src/calc/plugin.ron
+++ b/plugins/src/calc/plugin.ron
@@ -1,6 +1,6 @@
 (
     name: "Calculator",
-    description: "Math.JS calculations",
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
Thank you for updating this description.

friday:
> Thank you for updating this description. This does not affect the functionality of the calculator plugin.

True. I wasn't sure if the commit message should have been chore or fix, it changes the displayed name users see after all.

jacobgkau:
> it changes the displayed name users see after all.

So it does... I see now that the plugin descriptions are used in their help entries.

Screenshot from 2021-09-13 13-21-50

PR #27, opened 2021-09-22
Launcher Improvements

### LINKED ISSUE #19: Desktop entries deduplication
> If multiple files have the same desktop file ID, the first one in the $XDG_DATA_DIRS precedence order is used.
- https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html#desktop-file-id

Currently I have Nautilus duplicated in these two directories:
* `~/.local/share/applications/org.gnome.Nautilus.desktop`
* `/usr/share/applications/org.gnome.Nautilus.desktop`

I do this to override something in the desktop entries and avoid rewriting it when I update the package it belongs to.

My xdg_data_dirs: `~/.local/share`, `/usr/local/share` and `/usr/share` (`XDG_DATA_DIRS` is actually not set, but these are the paths that should be used when this is the case)

Screenshot from 2021-09-13 21-03-38

(same thing with "Bulk Rename", which is part of Thunar).

If you want this I could try to implement it, but I'm a Rust noob at best and haven't looked into your code.

### LINKED ISSUE #20: Support OnlyShowIn and NotShowIn
Desktop Entries allow apps to specify that they should show conditionally based on the Desktop Environment, using `OnlyShowIn` and `NotShowIn`

https://specifications.freedesktop.org/menu-spec/latest/ar01s03.html

I have XFCE installed and see those apps from running Gnome

Screenshot from 2021-09-13 21-21-00

The top one for example is /usr/share/applications/xfce4-power-manager-settings.desktop which specify `NotShowIn=GNOME;KDE;Unity;`

If you want this change I could try to implement it.

### LINKED ISSUE #23: Mentioning 'plugin.ron' is missing in README
I'm developing a basic plugin and found out that the plugin needs `plugin.ron` to be loaded by the launcher service. It doesn't seem to be mentioned in the README, though.

### LINKED ISSUE #25: Help does not show how to use features
Screenshot from 2021-09-19 08-30-22

mmstick:
Closes #19 
Closes #20 
Closes #23 
Closes #25

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1269,6 +1269,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "urlencoding",
+ "ward",
  "zbus",
  "zvariant",
 ]
@@ -1859,6 +1860,12 @@ name = "waker-fn"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9d5b2c62b4012a3e1eca5a7e077d13b3bf498c4073e33ccd58626607748ceeca"
+
+[[package]]
+name = "ward"
+version = "2.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1cbcd609d606e1307a1530061482ed2ee3fc9963128990238cefb2013127b61e"
 
 [[package]]
 name = "wasi"
--- a/README.md
+++ b/README.md
@@ -8,9 +8,35 @@ Using IPC enables each plugin to isolate their data from other plugin processes 
 
 ## Plugin Directories
 
-- User-local plugins: `~/.local/share/pop-launcher/plugins`
-- System-wide install for system administrators: `/etc/pop-launcher/plugins`
-- Distribution packaging: `/usr/lib/pop-launcher/plugins`
+- User-local plugins: `~/.local/share/pop-launcher/plugins/{plugin}/`
+- System-wide install for system administrators: `/etc/pop-launcher/plugins/{plugin}/`
+- Distribution packaging: `/usr/lib/pop-launcher/plugins/{plugin}/`
+
+## Plugin Config
+
+A plugin's metadata is defined `pop-launcher/plugins/{plugin}/plugin.ron`.
+
+```ron
+(
+    name: "PluginName",
+    description: "Plugin Description: Example",
+    bin: (
+        path: "name-of-executable-in-plugin-folder",
+    )
+    icon: Name("icon-name-or-path"),
+    // Optional
+    query: (
+        // Optional -- if we should isolate this plugin when the regex matches
+        isolate: true,
+        // Optional -- Plugin which searches on empty queries
+        persistent: true,
+        // Optional -- avoid sorting results from this plugin
+        no_sort: true,
+        // Optional -- pattern that a query must have to be sent to plugin
+        regex: "pattern"
+    )
+)
+```
 
 ## Script Directories
 
--- a/plugins/Cargo.toml
+++ b/plugins/Cargo.toml
@@ -29,3 +29,4 @@ tracing-subscriber = "0.2"
 urlencoding = "2"
 zbus = "1"
 zvariant = "=2.6" # Restrict for 1.47
+ward = "2.1.0"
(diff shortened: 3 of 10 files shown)


jackpot51 (approved):

jacobgkau:
After this, GNOME Control Center settings panels are no longer displaying in search results; those are part of the regression testing checklist. The G-C-C panels have `NoDisplay=true` set. Not sure what the best way to work around that would be.

mmstick:
We can ignore `NoDisplay` if `OnlyShowIn` was set. That was a condition the former logic was handling.

jacobgkau:
Regarding the new help copy, this is what it currently looks like:

Screenshot from 2021-09-23 10-32-46

I identified the following issues/limitations:
- The Web Search entry only shows DuckDuckGo as an example, with no way to discover the other 26 supported websites.
- The Calculator example (which is the only example that includes actual parameters instead of a placeholder) is a little difficult to understand, especially because there's an equals sign in the middle of the equation in addition to the prefix equals sign at the beginning.
  - The `(qalc)` just before the colon also makes this line look a bit messy, since the rest of the lines use similar-looking curly braces for required parameters.
- I'm not sure why the `OR` is in all caps in the file navigation example.

I tried playing around with making the usage patterns more consistent with man page format and having separate working examples, and came up with this:

Screenshot from 2021-09-23 12-22-52

This repurposes the name as a short description since there's not much value to having a non-descriptive short name listed. It doesn't solve the issue of plugins like web search that have too many options to list, and having a newline in the description seems like kind of a hack, but it might be an improvement overall. What do you think? (Wanted to check before I pushed; seems like UX might eventually need to be involved as well if this gets much more in-depth.)

WatchMkr:
@maria-komarova can you take a look at this when you have a chance?

On the one hand I like that not having as much detail makes it clean and has a good default for using the functionality. On the other hand, I can search youtube but don't know it and adding them while still looking clean is tricky.

maria-komarova:
General question before I dig deeper with this one - would having "Syntax" be useful? Because as a less technical user I find examples much more helpful. 
Overall I like shorter description, they are much easier to digest and grasp at a glance without spending too much time reading. Getting into details, I like the change from "Find" to "File search". But then longer descriptions on top of Syntax make it a little more difficult for me to quickly grasp the options. I'm also a bit confused on what "with tab autocomplete" means under File navigation.

jacobgkau:
> General question before I dig deeper with this one - would having "Syntax" be useful? Because as a less technical user I find examples much more helpful.

I think the reasons it's useful to include syntax would be:
- It makes more clear which words in the example should be changed (parameters) and which should stay the same (command/prefix).
- The syntax can have "or" operators `|` (or even just the word "or" like one has in the first image) to compactly show several different possible options that can be used, while doing the same thing with examples would require an entire example for every option.

> I'm also a bit confused on what "with tab autocomplete" means under File navigation.

That means when using file navigation, you can press tab to autocomplete directory/file names. Maybe should be changed to "with tab autocompletion" if not reworded further, but it is fairly useful to know about when using that plugin.

mmstick:
Shall I merge with the proposed changes from @jacobgkau ? We can always revisit the design of pop-shell later.

jacobgkau:
> Shall I merge with the proposed changes from @jacobgkau ? We can always revisit the design of pop-shell later.

I'd need to push the proposed changes if we want to do that, otherwise this is still in the state from my first screenshot. (Everything else did look good besides the help text that I brought up.)

mmstick:
Push to this branch if you can

maria-komarova:
I still think shorter descriptions are more user friendly when it comes to less technical users. I do like changing "Find" to "File Search" for the first line. And adjusting "OR" to "or" in the last "File navigation" row. 
I see the issues @jacobgkau pointed out but I can't see an easy way to improve it yet. More text makes it more difficult to scan through.

jacobgkau:
> I still think shorter descriptions are more user friendly when it comes to less technical users. I do like changing "Find" to "File Search" for the first line. And adjusting "OR" to "or" in the last "File navigation" row.

Okay, sounds like someone should revert that last commit and then make the two changes you're suggesting.

> More text makes it more difficult to scan through.

You don't think the labels at the front of the lines help with this? Having the syntax on the short lines at various points horizontally because of the different description lengths makes it more difficult to quickly find where those start, IMO, but you're the expert.

maria-komarova:
Let's push without the changes, even the ones I'm suggesting. I will make the design/description changes a separate thing. I'm sure there should be a better way of doing it somewhere there.

jackpot51 (approved):

jacobgkau:
@mmstick You just merged this with my two-line commit included.

maria-komarova:
It's fine then, let's keep two lines if they are already merged.

PR #42, opened 2021-10-11
feat(web): fetch and cache favicon

### LINKED ISSUE #41: [Feature request] - Per rule icon for web plugin
I would be nice to have an optional per rule icon for the web plugin, falling back to the default one when not specified :

```
        (
            matches: ["arch"],
            queries: [(name: "Arch Wiki", query: "wiki.archlinux.org/index.php/" )],
            icon: Name("archlinux")
        ),
        (
            // Should fallback to the icon defined in plugins/web/config.ron
            matches: ["bc", "bandcamp"],
            queries: [(name: "Bandcamp", query: "bandcamp.com/search?q=")]
        ),
```

oknozor:
There is bit too much unwrapping here but I am not sure everything needs to be checked. 
Let me know if any thing needs to be changed. 

Closes #41

### DIFF
--- a/plugins/src/web/mod.rs
+++ b/plugins/src/web/mod.rs
@@ -155,8 +155,15 @@ impl App {
 
             match icon {
                 Some(icon) => {
-                    let copy = smol::fs::write(&favicon_path, icon).await;
+                    // Ensure we recreate the pop-launcher cache dir if it was removed at runtime
+                    let cache_dir = favicon_path.parent().unwrap();
+                    if !cache_dir.exists() {
+                        smol::fs::create_dir_all(cache_dir)
+                            .await
+                            .expect("error creating cache directory");
+                    }
 
+                    let copy = smol::fs::write(&favicon_path, icon).await;
                     if let Err(err) = copy {
                         tracing::error!("error writing favicon to {:?}: {}", &favicon_path, err);
                     }
@@ -180,7 +187,6 @@ fn build_query(definition: &Definition, query: &str) -> String {
 
 async fn fetch_favicon(url: &str, favicon_path: &PathBuf, client: &HttpClient) -> Option<Vec<u8>> {
     let response = client.get_async(url).await;
-
     match response {
         Err(err) => {
             tracing::error!("error fetching favicon {}: {}", url, err);
@@ -228,14 +234,14 @@ async fn favicon_url_from_page_source(domain: &str, client: &HttpClient) -> Opti
                 if !icon_url.starts_with("https://") {
                     format!("https://{}{}", domain, icon_url)
                 } else {
-                    icon_url
+                    icon_url.into()
                 }
             }),
         Err(_err) => None,
     }
 }
 
-fn parse_favicon(mut html: &str) -> Option<String> {
+fn parse_favicon(html: &str) -> Option<String> {
     let regex = Regex::new(r"<!--(.+)-->").unwrap();
     let html = regex.replace_all(html, "").to_string();
 
@@ -254,7 +260,14 @@ fn parse_favicon(mut html: &str) -> Option<String> {
             let end = html.find("\"");
 
             if let Some(end) = end {
-                return Some(html[..end].to_string());
+                let icon_uri = &html[..end];
+                let icon_uri = if icon_uri.starts_with("//") {
+                    format!("https:{}", icon_uri)
+                } else {
+                    icon_uri.to_string()
+                };
+
+                return Some(icon_uri);
             }
         }
     }
@@ -264,9 +277,10 @@ fn parse_favicon(mut html: &str) -> Option<String> {
 
 #[cfg(test)]
 mod test {
-    use isahc::ReadResponseExt;
+    use isahc::{HttpClient, ReadResponseExt};
 
     use crate::web::parse_favicon;
+    use isahc::config::{Configurable, RedirectPolicy};
 
     #[test]
     fn should_parse_favicon_url_github() {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -62,6 +62,8 @@ pub enum IconSource {
     Name(Cow<'static, str>),
     // Icon is a mime type.
     Mime(Cow<'static, str>),
+    // Path to a cached icon.
+    Favicon(PathBuf),
```
mmstick, line 66: You can use `IconSource::Name` for this. It's generally assumed that if the name is a path (`starts with /`), file-based icon loading is performed.

oknozor: There is a 'static lifetime constraint, not sure I can use `IconSource::Name`. Any hint ?

mmstick:
Perhaps we should validate that the data we downloaded is an actual image file.

```sh
$ file --mime-type Search.ico
Search.ico: image/vnd.microsoft.icon
```
`image/png` and `image/gif` should also be valid outputs.

mmstick:
Be sure to update the debian/control to have a build-depend of `libssl-dev` on the source, and depend of `libssl1.1` on the package.

oknozor:
>This breaks bc/bandcamp search (the prefix no longer works), with this showing up in journalctl when I type that prefix:

Seems the problem is that bandcamp icon is located at https://s4.bcbits.com/img/favicon/favicon-16x16.png, The same goes for flathub and other broken endpoints, though I am not sure yet why the error message is different. 

I guess we should not expect other domains to always expose favicon at `{domain}/favicon.ico`. 
There is no default url scheme for favicons browser use the page source to determine the favicon location (https://www.w3.org/2005/10/howto-favicon)

That said, there is a google endpoint which could solve this for us : `https://www.google.com/s2/favicons?domain={domain}&sz={icon_size}`. I was not able to find any official documentation on this but is seems to be widely used. There are lots stackoverflow post and blogs article talking about this endpoint. 

> Attempting to search the Pop!_Planet forums with pp causes the launcher's results to stop responding to input until it's closed

This url https://pop-planet.info/favicon.ico, responds after one minute with a 504 Gateway Time-out. We should definitelty add a shorter time-out on our side.

>It could fetch the icon in the background (as a {filename}.partial) while returning the search result early. Then the launcher would have the option of showing no icon if it wasn't fetched in time.

@mmstick I am not familiar with smol but I gess it work quite like tokio tasks ? 

Tell me what you think about the google favicon endpoint, if its ok for you I will rewrite this as a background task querying for the favicon (if it is not cached already) via the google api.

oknozor:
I tried yo use scraper,  it uses Cell internally : 
```rust
        smol::spawn(async move {
            let response = client
                .get_async(url)
                .await;

            let html = response.unwrap().text().await.unwrap();
            let html = Html::parse_document(&html);
            let link_selector = Selector::parse(r#"link[rel="shortcut icon"]"#).unwrap();
            let link = html.select(&link_selector).next().unwrap();
            let icon_url = link.value().attr("href").unwrap();
// ... 
```

This produce the following error : 
```
error: future cannot be sent between threads safely
   --> plugins/src/web/mod.rs:145:9
    |
145 |         smol::spawn(async move {
    |         ^^^^^^^^^^^ future created by async block is not `Send`
    |
    = help: within `ego_tree::Node<Node>`, the trait `Sync` is not implemented for `Cell<NonZeroUsize>`
note: future is not `Send` as this value is used across an await
   --> plugins/src/web/mod.rs:183:32
    |
153 |             let link = html.select(&link_selector).next().unwrap();
    |                 ---- has type `ElementRef<'_>` which is not `Send`
...
183 |                     let copy = smol::fs::write(&favicon_path, contents).await;
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ await occurs here, with `link` maybe used later
...
190 |         })
    |         - `link` is later dropped here
note: required by a bound in `smol::spawn`
   --> /home/okno/.cargo/registry/src/github.com-1ecc6299db9ec823/smol-1.2.5/src/spawn.rs:30:67
    |
30  | pub fn spawn<T: Send + 'static>(future: impl Future<Output = T> + Send + 'static) -> Task<T> {
    |                                                                   ^^^^ required by this bound in `smol::spawn`

```

Maybe I can manually parse the href ? This would be trivial to implement I think.

oknozor:
>It should be fine if it is !Sync or !Send as long as its values aren't alive beyond an await point.

Oops I have just finished a scraper free implementation. Shall I try again with scraper ? 

Anyway, I have kept the google favicon as a fallback because google.com seems to set its favicon via javascript, making it impossible to get the icon from the html source.

jacobgkau:
`ahi`/`alie` queries are now showing a blank icon:

Screenshot from 2021-10-15 12-57-56

oknozor:
This is probably because the icon href start with a double slash on aliexpress : 
```html
<link rel="shortcut icon" type="image/x-icon" href="//ae01.alicdn.com/images/eng/wholesale/icon/aliexpress.ico" />
```
I will fix this tomorrow

oknozor:
@jacobgkau this should be working for all urls now :crossed_fingers:

jacobgkau:
While testing this earlier, I thought I saw the AliExpress icon working. However, now it's not working again, still on 7180df4. I'm getting this message in journalctl:
```
Oct 18 12:16:02 pop-os gnome-shell[17112]: Oct 18 12:16:02.375 ERROR pop_launcher_plugins::web: Got unexpected content-type 'text/html;charset=UTF-8' type for "/home/system76/.cache/pop-launcher/AliExpress.ico" favicon
```

Also, while troubleshooting this, I found that if I remove `~/.cache/pop-launcher/` and don't restart GNOME Shell, the web plugin stops working until I restart GNOME Shell:
```
Oct 18 12:17:05 pop-os gnome-shell[17551]: thread 'main' panicked at 'unable to create $HOME/.cache/pop-launcher: Os { code: 2, kind: NotFound, message: "No such file or directory" }', plugins/src/web/mod.rs:65:41
```

jacobgkau:
> ali express uses a `//` prefix for the favicon href, its an url without the protocol

No worries, my understanding is that `//` means "use whatever protocol was used to access the webpage," so a browser will use http:// or https:// depending on whether the user connected to the site via http:// or https://.

> Best to go with a crate that has figured HTML parsing out.

I was just minutes away from approving this, all icons are working now, but we can hold off if you'd rather this be done differently.

PR #63, opened 2021-11-18
feat(plugin-pop-shell): Get window icon from desktop entry

mmstick:
Requires https://github.com/pop-os/shell/pull/1246


jackpot51 (commented):
Can probably remove IconSource::Window

jackpot51 (approved):

jacobgkau (approved):
Tested with https://github.com/pop-os/shell/pull/1246.

PR #65, opened 2021-11-24
IPC Client API

mmstick:
- Relicense crate libraries to MPL-2.0 to permit Cargo linking while preserving copyleft requirements
- Plugins + Binary crate are to remain as GPL-3.0
- Add IPC Client API for future launcher frontend.

Makes no changes to existing behavior of the launcher service used by pop-shell.


jackpot51 (approved):

jacobgkau (approved):
Launcher regression testing passed & all plugins still work. Tested on 21.10 and 20.04.

This branch does not seem to include https://github.com/pop-os/launcher/pull/66 yet, but does not appear to undo that PR based on the diff.

PR #66, opened 2021-11-24
chore(plugins-web): Change aliases to address complaints

mmstick:
People don't like that `a` features Amazon and `b` features Bing. I'm
sure the same could be said for `g` for Google. Single-character aliases
seem to be a bad idea. Other aliases could also use some improvements
to add additional obvious aliases.

https://www.reddit.com/r/pop_os/comments/r19195/pop_os_how_do_i_stop_the_shell_from_showing/

### DIFF
--- a/plugins/src/web/config.ron
+++ b/plugins/src/web/config.ron
@@ -5,7 +5,7 @@
             queries: [(name: "AliExpress", query: "aliexpress.com/wholesale?SearchText=")]
         ),
         (
-            matches: ["a", "amazon"],
+            matches: ["amazon"],
             queries: [(name: "Amazon", query: "smile.amazon.com/s?k=" )]
         ),
         (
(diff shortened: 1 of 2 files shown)


jackpot51 (approved):

jackpot51:
I would recommend pinning a comment on each reddit thread pointing to this.

jackpot51 (changes requested):
I think the help messages need to be updated

K-e-d-i:
I would like to propose, on top of what @mmstick suggested, to add a toggle for disabling web searches from the launcher and make this toggle persist through updates.

mmstick:
@K-e-d-i What you're really asking for is just a way of enabling and disabling plugins. There isn't currently a configuration mechanism for that, but it's planned.

K-e-d-i:
@mmstick  Ahh, if that is the case, +1 for that future plan. Current solution to disable this feature is less than user friendly in my opinion.

jacobgkau (approved):
This successfully changes the single-character prefixes to multi-character, and adds a few new longer prefixes. The help entry's text matches the new prefix for Google.

PR #74, opened 2021-12-20
feat(desktop-entries): Nix application paths support

mmstick:
The Nix system path takes priority over the system path, but below Flatpak and Snap system paths, and the local nix path takes priority over all system paths, but below local Flatpak/Snap/.local paths.

To install nix in a multi-user capacity (system-wide):

```
sh <(curl -L https://nixos.org/nix/install) --daemon
```

Then you can install an application per-user and see it installed at `~/.nix-profile/share/applications`.

```
nix-env -i firefox
```

To install system-wide, switch to root and do the same and you'll find it installed in `/nix/var/nix/profiles/default/share/applications`

```
sudo su
bash
nix-env -i firefox
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -540,9 +540,9 @@ dependencies = [
 
 [[package]]
 name = "freedesktop-desktop-entry"
-version = "0.4.0"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "61ecb2f63bfff12ca1a6e98c2ed7bac47169abe4d3cdbaeeb8e3852afdcc8ace"
+checksum = "bf77a18f5a7dc4fa975cb54cbb204abc04121131a92aa7010438d29e513bb1ae"
 dependencies = [
  "dirs",
  "gettext-rs",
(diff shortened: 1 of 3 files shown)


jackpot51 (approved):

jacobgkau:
> The Nix system path takes priority over the system path, but below Flatpak and Snap system paths, and the local nix path takes priority over all system paths, but below local Flatpak/Snap/.local paths.

After I installed Firefox with Nix, I still see the Flatpak option below it in search results (which seems correct), but I don't see the system-installed Firefox in the results anymore:

Screenshot from 2021-12-21 12-37-29

Is this expected? I would think either I should see all of the Firefox versions (which I'd prefer), or else only one.

mmstick:
It is expected behavior because the Flatpak has a differing desktop ID from the debian and nix versions of it. The Flatpak uses `org.mozilla.firefox` and both the debian and nix versions use `firefox`.

jacobgkau (approved):
This is working as expected. I see `Nix` and `Nix (System)` entries in the launcher, with local `Nix` entries overriding `Nix (System)` entries, similar to Flatpak.

Launcher regression testing passed on 21.10 and 20.04.

PR #75, opened 2021-12-21
fix(pop_shell): Implement 'Quit' request handler

ningvin:
This addresses [an issue opened in the pop-os/shell repo](https://github.com/pop-os/shell/issues/1278) (is there a more convenient way to link to issues of a different project...?), where the ability to close an item in the launcher window via `Ctrl`+`Q` was lost, presumably after replacing the JS service with the Rust service.

The request handler for the *Quit* event of the *pop_shell* plugin just did nothing, instead of calling the `WindowQuit` function via D-Bus.

I am assuming this is just an oversight and there are no technical limitations preventing this (it seemed to work just fine when I tested it locally), hence this pull request :-)

### DIFF
--- a/plugins/src/pop_shell/mod.rs
+++ b/plugins/src/pop_shell/mod.rs
@@ -45,7 +45,7 @@ pub async fn main() {
         match request {
             Ok(request) => match request {
                 Request::Activate(id) => app.activate(id).await,
-                Request::Quit(_id) => (),
+                Request::Quit(id) => app.quit(id).await,
                 Request::Search(query) => app.search(&query).await,
                 Request::Exit => break,
                 _ => (),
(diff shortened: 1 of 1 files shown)


mmstick (approved):

13r0ck (approved):
I didn't know the launcher used to do this. Very useful. Thank you for fixing it!


Also, to answer your question you can link issued in other repos like this:
\<org>/<repo>#<issue/pr number> and git will auto-generate the link.
`pop-os/shell#1278` -> pop-os/shell#1278

jacobgkau (approved):
Tested with https://github.com/pop-os/shell/pull/1286.

ningvin:
> I didn't know the launcher used to do this. Very useful. Thank you for fixing it!
> 
> Also, to answer your question you can link issued in other repos like this: <org>/#<issue/pr number> and git will auto-generate the link. `pop-os/shell#1278` -> [pop-os/shell#1278](https://github.com/pop-os/shell/issues/1278)

Glad I could help and thanks for the info :)

PR #80, opened 2022-01-11
fix(terminal): Do not invoke read command on shell execution

mmstick:
This will prevent a lingering process when executing `:command` in the launcher

### DIFF
--- a/plugins/src/terminal/mod.rs
+++ b/plugins/src/terminal/mod.rs
@@ -44,8 +44,8 @@ pub async fn main() {
 
 impl App {
     async fn activate(&mut self, _id: u32) {
-        let exec = match self.last_query.take() {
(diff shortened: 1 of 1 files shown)


jacobgkau:
On 60a4e2a, I'm still seeing an extra process show up and remain in the list every time I run a command with the `:` prefix. (I'm reading the list with `px aux | grep launcher`.) Is it still expected for these processes to keep showing up after this fix?

mmstick:
That should be fixed now

jacobgkau (approved):
The fix is working now.

Regression tested the Terminal plugin's three prefixes (the two foreground prefixes still show the `Press Enter to exit` message as expected); this plugin is the only file being modified by this PR.

PR #91, opened 2022-03-04
fix: Show context for graphics only if 2+ graphics devices found

### LINKED ISSUE #89: Right click "launch using discrete graphics" pop-up appears on machines with only integrated graphics
This was observed on lemp9. Strangely, it does not appear on my oryp6 while in Nvidia mode.

mmstick:
Closes #89


leviport (approved):
This hides the context on integrated-only machines and on switchable machines in Integrated mode. I still see it on switchable machines in Nvidia mode, but I'm approving since it's still an improvement.

PR #97, opened 2022-03-29
Fixed web plugin favicons, binary size reduction, perf optimizations

### LINKED ISSUE #98: Desktop entries plugin: unknown cmd
I'm running version 1.2.0 (commit bacada02c23ab7df021f8b8ead7470df09443cb5) on an Arch machine, and I'm seeing no desktop entries in the launcher GUI.

After some investigation, running the `desktop_entries` plugin doesn't work, the binary spits out the following error:
```
unknown cmd: pop-launcher-bin
```
and exits without any error codes.

Reproduced both when installed from AUR and on a fresh clone.

<details>
  <summary>Reproduce steps</summary>

  ```console
  $ just
  ...
  $ cd target/release
  $ ln -s pop-launcher-bin desktop_entries
  $ ./desktop_entries
  unknown cmd: desktop_entries
  $ echo $?
  0
  ```

</details>

mmstick:
- Enable LTO, panic=abort, set strip-true by default to reduce binary size
- Swap the default system allocator for mimalloc, which is 2.6x faster and more secure
- Fixed the symbolic link names installed by the justfile
- Fixed missing favicons on some websites, like Amazon
    - It seems Google isn't correctly checking the root of a domain for the favicon anymore

Binary size should now be reduced from ~8MB to ~6MB, which is shared memory to all plugins.

Closes #98


jacobgkau:
This fixes favicon caching for Amazon and Crates.io. However, it breaks the icon for Flathub.

Screenshot from 2022-03-29 11-07-00

Interestingly, the DEV icon has changed. Before, it used the same favicon that Firefox recognizes:

Screenshot from 2022-03-29 11-08-53

After, it's different:

Screenshot from 2022-03-29 11-07-03

mmstick:
Seems to be the icon they're using here: https://dev.to/favicon.ico

jacobgkau:
> Seems to be the icon they're using here: https://dev.to/favicon.ico

They have their `shortcut icon` set to a `favicon.ico` on a different domain:

image

jackpot51:
- Swap reqwest for ureq for significant reduction in dependencies and binary size

I do not recommend this. I have noticed reqwest has better performance than ureq.

mmstick:
I've reverted the change to keep reqwest

jackpot51 (approved):

jacobgkau:
Amazon and Crates.io are showing the Google API's default favicon instead of their own icons again on 09d126e.

mmstick:
Seems the correct order would be first scraping for the favicon, then checking the root path, and falling back on Google as a last resort.

jacobgkau:
On [ca4eec7](https://github.com/pop-os/launcher/pull/97/commits/ca4eec72a6b07caca6372cbd8864d248133bef98), the lib.rs and ArXiv icons don't load:

Screenshot from 2022-03-30 09-23-17

Screenshot from 2022-03-30 09-25-36

The dev icon doesn't load, and doesn't show a placeholder space either, and it shows a backtrace:

Screenshot from 2022-03-30 09-23-50

jacobgkau (approved):
On e5f381b, all favicons are showing as expected again. Regression testing passed:
[launcher-pr97-impish.txt](https://github.com/pop-os/launcher/files/8382540/launcher-pr97-impish.txt)
[launcher-pr97-focal.txt](https://github.com/pop-os/launcher/files/8382611/launcher-pr97-focal.txt)

PR #109, opened 2022-05-11
fix(plugin): load desktop entries from user dir

oknozor:
Currently the destkop-entry plugin retains the entries emitted by the freedesktop-desktop-entry crate in reverse order. 

For instance '$HOME/.local/share/applications/jetbrains-clion.desktop' is emitted first then 
overriden by '/usr/share/applications/jetbrains-clion.desktop'.  

Instead of `Hashset::replace`, `Hashset::insert` must be use to avoid overriding entries with higher priority. 

From freedesktop basedir specification: 
- " There is a set of preference ordered base directories relative to which data files should be searched. This set of directories is defined by the environment variable $XDG_DATA_DIRS." (see: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

From desktop-entries specification: 
-  "If multiple files have the same desktop file ID, the first one in the $XDG_DATA_DIRS precedence order is used." (see: https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html)

### DIFF
--- a/plugins/src/desktop_entries/mod.rs
+++ b/plugins/src/desktop_entries/mod.rs
@@ -168,7 +168,7 @@ impl<W: AsyncWrite + Unpin> App<W> {
                                 src,
                             };
 
-                            deduplicator.replace(item);
+                            deduplicator.insert(item);
                         }
                     }
                 }


mmstick (approved):

jacobgkau (approved):
Confirmed this gets local entries in `~/.local/share/applications/` overriding system entries in `/usr/share/applications/` as expected.

I had some trouble while testing this; first, when I copied  `/usr/share/applications/org.gnome.gedit.desktop` to `~/.local/share/applications/org.gnome.gedit.desktop`, the entry showed up as `Local` in the launcher and changes to the name/icon worked, but changes to the `Exec=` lines did not take effect (after changing them all to `gnome-calculator`, for example, it still launched Gedit.) However, the changes also didn't apply when launching from the Applications menu, so it didn't seem like a pop-launcher issue. I then tried changing the original file at `/usr/share/applications/org.gnome.gedit.desktop`, and changes to its `Exec=` lines didn't apply either.

Next, I tried using Firefox to test instead, in case GNOME was doing something special with Gedit. For Firefox, I copied its .desktop file to the local directory and changed its `Exec=` lines to say `gnome-calculator`, and that change worked. However, after going back to the current stable pop-launcher (causing the entry to say `System` again instead of `Local`), the `System` entry continued launching Calculator instead of Firefox.

I'm approving because this improves behavior (makes locally-overridden items show up as `Local` and gets name/icon changes working, at least.) If pop-launcher uses a GNOME mechanism to launch .desktop files instead of running the `Exec=` line directly, that would explain some of the discrepancies I'm seeing with the current version launching local overrides correctly without this change. One other thing I noticed was that `/home/$USER/.local/share` is not part of `$XDG_DATA_DIRS` by default on Pop, but applications from `~/.local/share/applications` show up without it.

PR #112, opened 2022-05-11
feat: add a history attribute to plugin config as described in #110

### LINKED ISSUE #110: Introduce an optional "history" attribute in plugin config.
Hello,

I am working on onagre these days, and I was wondering if you would accept adding an optional `history` attribute to plugin config. 
The idea is to indicate that the client must display user history for the plugin when the plugin gets activated but no request was send to pop-launcher yet. 

## Example : 

**config:**
```ron
(
    name: "Desktop Entries",
    description: "Query applications by their .desktop entries",
    bin: (path: "desktop-entries"),
    icon: Name("new-window-symbolic"),
    history: true
)
```
When there is not user input the client should display history for the current plugin. It would be the client responsibility to 
store the history and launch these entries ( currently onagre just send and `Search` request followed by `Acivate(0)`). 

image

Use pop-launcher response when it's available : 
image

## Motivation 

History is useful for some plugins like `desktop-entry`, `web`, `terminal`, `calc`, on the other hand, displaying a history for `files`, `recent` does not make sense. 

This was already implemented in onagre but I am now trying to remove all the custom external command stuff and rely on exclusively pop-launcher plugins configuration. 

Without this flag directly available in the plugin config I need to define some external configuration to get the desired behavior for a plugin. This means forcing the user to maintain a separate config and edit this whenever they install a new plugin. 

I don't now I this would be useful for cosmic and other available client . Maybe you already handle this in a different way ? 

Let me know if this would be ok for you and I'll send a PR.

oknozor:
closes #110

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ A plugin's metadata is defined `pop-launcher/plugins/{plugin}/plugin.ron`.
     description: "Plugin Description: Example",
     bin: (
         path: "name-of-executable-in-plugin-folder",
-    )
+    ),
     icon: Name("icon-name-or-path"),
     // Optional
     query: (
@@ -65,7 +65,9 @@ A plugin's metadata is defined `pop-launcher/plugins/{plugin}/plugin.ron`.
         // Optional -- avoid sorting results from this plugin
         no_sort: true,
         // Optional -- pattern that a query must have to be sent to plugin
-        regex: "pattern"
+        regex: "pattern",
+        // Optional -- the launcher should keep an history for this plugin
+        history: true,
     )
 )
 ```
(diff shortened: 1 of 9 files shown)


mmstick (approved):

jacobgkau (approved):
Launcher regression testing passed. [launcher-pr112-jammy.txt](https://github.com/pop-os/launcher/files/8673622/launcher-pr112-jammy.txt)

This does not seem to affect any behavior with the pop-shell frontend (for example, the web plugin does not show search history after the prefix has been typed but before the rest of the query, even though it now has `history: true,` in its `plugin.ron` file.)

mmstick (approved):

PR #116, opened 2022-05-25
add UPS search query

ahoneybun:
This PR does the following:

- Add UPS tracking number search


mmstick (dismissed):

jacobgkau:
This is a great idea. It seems like the `loc=null` parameter isn't needed, is there a reason you included that? @ahoneybun

ahoneybun:
No reason I just copied it from a base URL, if it can be removed and still work I'm fine with that.

mmstick (approved):

PR #119, opened 2022-05-27
Feat/add desktop entries action

oknozor:
This add the ability to use [desktop entries actions](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#extra-actions). 

Not sure but, this might break pop-shell front end.  Maybe I should add another variant to the `PluginResponse` instead of using `Context` ? 

Also I am wondering if we should implement activation for destkop entries directly in the launcher or still delegate to clients ?  (see: https://github.com/pop-os/launcher/discussions/118).


oknozor:
I think it would be better to have desktop entries ans actions activated by pop-launcher as mentioned here https://github.com/pop-os/launcher/discussions/118

I will try to tackle this soon, and make  a PR in pop-shell in parallel.

oknozor:
I updated the desktop entries plugin so it now take care of launching the entries, actions and gpu preferences.

Still need to figure out how to properly handle GPU selection for nvidia though. I don't have an nvidia machine so I won't be able to test this. 

Should I implement something like this: https://gitlab.freedesktop.org/hadess/switcheroo-control/-/blob/master/src/switcheroo-control.c#L285 ?  

Could you point me in the right direction @mmstick. 

Also this will require to remove all the desktop entry related code in pop-shell. I'll send the PR soon.

PR #123, opened 2022-06-04
Add Ecosia Search to config.ron

HarshAsrani:
Added Ecosia Search as an option in the launcher


n3m0-22 (changes requested):
Looks good and is working as expected.

> Suggest adding "eco" as an additional shorthand match; otherwise, this looks good.

I agree with @Affebonbons that it would be beneficial to have a shorthand.

n3m0-22 (approved):

13r0ck (approved):
Thank you!

Issue #126, opened 2022-06-26
Calculator ignores equations beginning with special characters


titojankowski:
Distribution (run cat /etc/os-release):
22.04

Issue/Bug Description:
Starting a calculation with a period or parentheses doesn't result in a calculation

Steps to reproduce (if you know):

Example A:
- Start the Launcher
- Enter a calculation that begins with a period, such as ".1+1"
- No result appears

Example B:
- Start the Launcher
- Enter a calculation that begins with a parentheses, such as "(1+1)"
- No result appears

Expected behavior:
Example A:
Calculator should show 1.1 as the top result

Example B:
Calculator should show 2 as the top result

Other Notes:
Related to #47

titojankowski:
Are there any other valid "start" characters besides "." and "(" for an equation?

Negative sign "-" works.

PR #127, opened 2022-06-27
addition to regex

n3m0-22:
This addition fixes https://github.com/pop-os/launcher/issues/126


gangwerz (approved):
- `.1+1` shows the correct result of `1.1`
- `(.1+1)` shows the correct result of `1.1` as well

PR #128, opened 2022-06-28
fix(web): Prevent overwriting of keywords

### LINKED ISSUE #125: Web Plugin User Config is not prioritized
Writing a user specific web plugin config to `.local/share/pop-launcher/plugins/web/config.ron` adds new queries to the webplugin but does not take priority over existing queries.

Example:  
I added a few queries to my `.local`-plugin folder.
While new queries like for example
```ron
(
   matches: ["np"],
   queries: [(name: "Nix Packages", query: "search.nixos.org/packages?channel=unstable&query=")]
),
```
are added successfully, queries like:
```ron
(
   matches: ["gn"],
   queries: [(name: "GitHub Nix Code", query: "github.com/search?type=Code&l=Nix&q=")]
),
```
are shadowed by the config provided in the DISTRIBUTION path.

I feel like the user config should always take priority over the system-wide config and the system-wide config should always take priority over the distribution config.

Steav005:
While the configs come in, in the correct order (User, System, Distro), inside of the web plugin, keywords are overwritten by later keywords. This prevents the overwriting of already present keywords.

Follows clippy lint convention: https://rust-lang.github.io/rust-clippy/master/#map_entry

Fixes #125

### DIFF
--- a/plugins/src/web/config.rs
+++ b/plugins/src/web/config.rs
@@ -16,7 +16,7 @@ impl Config {
         for rule in rules.rules {
             let idx = self.queries.insert(rule.queries);
             for keyword in rule.matches {
-                self.matches.insert(keyword, idx as u32);
+                self.matches.entry(keyword).or_insert(idx as u32);
             }
         }
     }


mmstick (approved):

n3m0-22 (approved):
This looks good. I tested with the examples in https://github.com/pop-os/launcher/issues/125 both are working. Using `gn` now searches GitHub Nix Code instead of Google News. I was able to switch other defaults as well with no issue. All other launcher functionality is working with no regressions.

PR #129, opened 2022-06-29
feat(web): specify icon

Steav005:
Some favicons are only provided with a resolution of 16x16, resulting in the logo being rather muddy inside of pop-launcher.

WIth this pull-request some changes were made.
- The Definition-Struct was extended with the icon-variable, which does not need to be defined.
  - If it is defined, it is prioritized as a candidate for the downloaded icon.
- The cascading fallback icon download match-cases were replaced with a List of icon sources.
  - The first successful icon of this list is used. 

As an example Nix Packages can be used
```
(
   matches: ["np"],
   queries: [(name: "Nix Packages", query: "search.nixos.org/packages?channel=unstable&query=")]
),
```
With this a rather muddy looking icon is used.
But should we use this instead:
```
(
   matches: ["np"],
   queries: [(
      name: "Nix Packages", 
      query: "search.nixos.org/packages?channel=unstable&query=",
      icon: "https://nixos.wiki/nixos-logo-small.png",
   )]
),
```
A more crisp looking icon is used instead.

I am not to proud of the for-loop, processing the icon_sources-list.
The alternative would be to convert the icon_sources-list to a [futures stream](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html), but this introduced some problems and made it less obvious what is happening.

### DIFF
--- a/plugins/src/web/config.rs
+++ b/plugins/src/web/config.rs
@@ -44,6 +44,8 @@ pub struct Rule {
 pub struct Definition {
     pub name: String,
     pub query: String,
+    #[serde(default)]
+    pub icon: String,
 }
 
 pub fn load() -> Config {
--- a/plugins/src/web/mod.rs
+++ b/plugins/src/web/mod.rs
@@ -12,7 +12,7 @@ use url::Url;
 
 use pop_launcher::*;
 
-pub use config::{Config, Definition, load};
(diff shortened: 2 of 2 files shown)


n3m0-22 (approved):
Looks good. This makes a nice improvement to the icon. No regressions.

Before:
before

After:
after

PR #134, opened 2022-08-03
Add Fedex tracking

ahoneybun:
This PR does the following:

- Adds Fedex tracking to launcher


Absolucy (approved):
No clue why one of the jenkins builds is being reported as a failure, but when I view it in jenkins, it's a success.

Looks simple and fine enough to me, though.

n3m0-22 (approved):
Tested with an old tracking number that still shows history. This is working great.

PR #140, opened 2022-09-03
add startpage to config.ron

Chinchzilla:
I was missing starpage in pop launcher so I added it into 'launcher/plugins/src/web/config.ron' with either prompt 'sp' or 'startpage'.


mmstick (dismissed):

n3m0-22 (approved):
This is working correctly. Thank you for contributing.

13r0ck (approved):

PR #144, opened 2022-09-29
Add poweroff keyword to session-shutdown.sh and sleep keyword to session-suspend.sh

merelymyself:
`poweroff` is the command used in the terminal, and `sleep` will be what Windows users will likely be used to. Apologies if this PR is unnecessary.


mmstick (approved):

n3m0-22 (approved):
This is working as expected. Looks good to me. Thank you contributing.

PR #145, opened 2022-10-03
Add Google Scholar to the search engines.

BramVerb:
I would like to search google scholar through the pop os launcher. The matches are subject to discussion, I typically use `scholar`, but DuckDuckGo uses the `!gsc` shortcut, so I added that as well.

### DIFF
--- a/plugins/src/web/config.ron
+++ b/plugins/src/web/config.ron
@@ -80,7 +80,7 @@
             queries: [(name: "News", query: "google.com/news?q=")]
         ),
         (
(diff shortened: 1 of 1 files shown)


BramVerb:
I also noticed that the indentation of all entries is not consistent, if you want I can also fix that.
Most uses spaces, but some use tabs. 😅

n3m0-22 (approved):
This is working correctly. Thank you for your contribution.

mmstick:
Thanks!

mmstick (approved):

PR #147, opened 2022-10-30
fix: systemd fallbacks for session scripts

jokeyrhyme:
### Fixed

- added fallbacks in session scripts for systems that do not have GNOME but do have systemd

### DIFF
--- a/scripts/session/session-logout.sh
+++ b/scripts/session/session-logout.sh
@@ -1,8 +1,20 @@
-#!/bin/sh
+#!/usr/bin/sh
 #
 # name: Log Out
 # icon: system-log-out
 # description: Log out to the login screen
 # keywords: log out logout
 
-gnome-session-quit --logout
+set -eu
+
+is_gnome() {
+  command -v dbus-send >/dev/null && \
+    command -v gnome-session-quit >/dev/null && \
+    dbus-send --print-reply --dest=org.gnome.Shell /org/gnome/Shell org.freedesktop.DBus.Properties.Get string:org.gnome.Shell string:ShellVersion >/dev/null 2>&1
+}
+
+if is_gnome; then
+  gnome-session-quit --logout
+elif command -v loginctl >/dev/null; then
+  loginctl terminate-user "${USER}"
+fi
--- a/scripts/session/session-reboot.sh
+++ b/scripts/session/session-reboot.sh
@@ -1,8 +1,20 @@
-#!/bin/sh
+#!/usr/bin/sh
 #
 # name: Restart
 # icon: system-restart
 # description: Reboot the system
 # keywords: power reboot restart
 
-gnome-session-quit --reboot
+set -eu
+
+is_gnome() {
+  command -v dbus-send >/dev/null && \
+    command -v gnome-session-quit >/dev/null && \
+    dbus-send --print-reply --dest=org.gnome.Shell /org/gnome/Shell org.freedesktop.DBus.Properties.Get string:org.gnome.Shell string:ShellVersion >/dev/null 2>&1
+}
+
+if is_gnome; then
+  gnome-session-quit --reboot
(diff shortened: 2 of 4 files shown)


scripts/session/session-logout.sh
```diff
@@ -1,8 +1,14 @@
...
 #
 # name: Log Out
 # icon: system-log-out
 # description: Log out to the login screen
 # keywords: log out logout
 
-gnome-session-quit --logout
+set -eu
+
+if command -v gnome-session-quit >/dev/null; then
```
mmstick, line 10: This would trigger if someone had GNOME installed even if they weren't actually in GNOME

mmstick (changes requested):

scripts/session/session-logout.sh
jokeyrhyme: True, but it's not worse than the status quo in that regard

I'm not entirely sure which environment variables we reliably have access to within these scripts

I did think about checking if XDG_CURRENT_DESKTOP=GNOME but didn't want to risk breaking this for GNOME users if that guess was wrong

I'll install GNOME and see if I can improve this PR, cheers

jokeyrhyme:
Okay, I've installed GNOME and confirmed that `if [ "${XDG_CURRENT_DESKTOP:-}" = "GNOME" ] && command -v gnome-session-quit >/dev/null; then` works as expected when in a GNOME session that is started by GDM, which I assume is the common case for GNOME sessions

I've updated the PR accordingly

mmstick (commented):
We can use dbus-send to check if gnome-shell is running.

mmstick:
```sh
is_gnome() {
    dbus-send --dest=org.gnome.Shell /org/gnome/Shell org.freedesktop.DBus.Properties.Get string:org.gnome.Shell string:ShellVersion
}

if is_gnome; then
```

jokeyrhyme:
Huh, I have to add the `--print-reply` flag to get that to return with an exit code when GNOME is not running:

```
❯ dbus-send --print-reply --dest=org.gnome.Shell /org/gnome/Shell org.freedesktop.DBus.Properties.Get string:org.gnome.Shell string:ShellVersion
Error org.freedesktop.DBus.Error.ServiceUnknown: The name org.gnome.Shell was not provided by any .service files
```

On my Archlinux system, it looks like `/usr/bin/dbus-send` is provided by the dbus package, which is required by gtk2, gtk3, pipewire, and systemd, but doesn't seem directly required by any of the GNOME packages (not even gtk4, weirdly), but it seems safe to assume that most GNOME systems will have it

scripts/session/session-logout.sh
```diff
@@ -1,8 +1,21 @@
...
 # name: Log Out
 # icon: system-log-out
 # description: Log out to the login screen
 # keywords: log out logout
 
-gnome-session-quit --logout
+set -eu
+
+is_gnome() {
+  [ "${XDG_CURRENT_DESKTOP:-}" = "GNOME" ] && \
```
mmstick, line 11: This fails on Pop!_OS because XDG_CURRENT_DESKTOP is `Unity`. Only the dbus command is required to check if gnome-shell is running.

mmstick (changes requested):

jokeyrhyme:
Okay, I _think_ we've got this down to just what is needed and nothing more :)

mmstick (approved):

jokeyrhyme:
@merelymyself I can see you've contributed to these scripts in the past, would you care to review this PR (completely understand if you'd rather not)? :smile:

mmstick:
The only review that matters right now is QA

n3m0-22 (approved):
Looks good to me. I'm not seeing any regressions. Thanks for your contribution.

jokeyrhyme:
Thanks all <3

PR #148, opened 2022-11-04
Dependency updates

mmstick:


service/src/lib.rs
```diff
@@ -295,8 +300,12 @@ impl<O: futures::Sink<Response> + Unpin> Service<O> {
     async fn activate_context(&mut self, id: Indice, context: Indice) {
         let mut ex = None;
         if let Some((plugin, meta)) = self.search_result(id as usize) {
+<<<<<<< HEAD
```
13r0ck, line 303: it looks like there are some improperly resolved git conflicts here

13r0ck (changes requested):

13r0ck (approved):

PR #155, opened 2023-01-11
fix: allow web plugin to xdg-open custom URI schemes #154

### LINKED ISSUE #154: Launch an xdg-open custom URI
The https://obsidian.md app (flatpak) registers its own mime type, `obsidian://...`. From the CLI, I can launch Obsidian with an open search page, like so:

```
$ xdg-open obsidian://search?query=journal
```

I can add a launcher config entry, like so:

```
$ sudo vim /usr/lib/pop-launcher/plugins/web/config.ron

#
#    (
#       matches: ["j", "journal"],
#       queries: [(name: "Journal", query: "obsidian://search?query=")]
#    ),
#
```

However, the launcher seems to open a browser to `https://obsidian://search?query=`.

Is this intended behavior?

canadaduane:
Prevents prefixing `https://` to custom URI schemes

Fixes #154

### DIFF
--- a/plugins/src/lib.rs
+++ b/plugins/src/lib.rs
@@ -13,7 +13,7 @@ pub mod terminal;
 pub mod web;
 
 use pop_launcher::PluginResponse;
-use std::{borrow::Cow, ffi::OsStr, future::Future, path::Path};
+use std::{borrow::Cow, ffi::OsStr, future::Future, path::Path, process::Stdio};
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 
 pub async fn send<W: AsyncWrite + Unpin>(tx: &mut W, response: PluginResponse) {
@@ -47,5 +47,9 @@ pub fn mime_from_path(path: &Path) -> Cow<'static, str> {
 
 /// Launches a file with its default appplication via `xdg-open`.
 pub fn xdg_open<S: AsRef<OsStr>>(file: S) {
-    let _ = tokio::process::Command::new("xdg-open").arg(file).spawn();
+    let _ = tokio::process::Command::new("xdg-open")
+        .arg(file)
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .spawn();
 }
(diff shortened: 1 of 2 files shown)


mmstick (approved):

n3m0-22 (changes requested):
@canadaduane Is this working on your end? I've never been able to get it working in my testing. I created a vault in Obsidian called journal, and added the new entry in the `config.ron` file. The regex does not seem to not be working correctly. I have tried a few different variation on it, but no luck.

Note: I am adding the new entry to `~/.local/share/pop-launcher/plugins/web/config.ron` so it is not lost between updates.

canadaduane:
> The regex does not seem to not be working correctly.

Can you provide any detail on this? For example, does it prevent you from opening `http` or `https` URLs? It should continue to work for `http` and `https`, and of course also allow new schemes such as `obsidian:`.

Regarding Obsidian's registered mime-type and xdg-open specifically, I've documented that I cannot get it to work yet [here](https://github.com/pop-os/launcher/issues/154#issuecomment-1378213759) due to a different issue.

n3m0-22 (dismissed):
You're correct it does still work with URLs, and does allow for schemes like `obsidian:`. That's my fault I swapped the timestamps in my head on this PR and [this comment](https://github.com/pop-os/launcher/issues/154#issuecomment-1378213759). I had thought this PR came after, and that the issue had been resolved.

In that case this is working as expected, and I am seeing no regressions. I see no reason not to approve.

Thank you for the clarification, and thank you for contributing to Pop!_OS.

mmstick:
May just require tweaking the [xdg_open](https://github.com/pop-os/launcher/blob/5adac084ffa7c545b53be4d89a22740b66860507/plugins/src/lib.rs#L49) function to set stdout and stderr to null

canadaduane:
> May just require tweaking the [xdg_open](https://github.com/pop-os/launcher/blob/5adac084ffa7c545b53be4d89a22740b66860507/plugins/src/lib.rs#L49) function to set stdout and stderr to null

You're right! It works. I'll add this to the PR.

n3m0-22 (approved):
Looks good. Working as expected.

mmstick (approved):

PR #156, opened 2023-01-12
Add generic `www` pattern to open any website

canadaduane:
Opens a website in default browser, e.g. `www github.com`

### DIFF
--- a/plugins/src/web/config.ron
+++ b/plugins/src/web/config.ron
@@ -132,6 +132,10 @@
             queries: [(name: "Wikipedia", query: "wikipedia.org/w/index.php?search=")]
         ),
         (
+            matches: ["www"],
+            queries: [(name: "Open Website", query: "http://")]
+        ),
+        (
             matches: ["xaut", "arxauth", "arxivauthor"],
             queries: [(name: "ArXiv", query: "https://arxiv.org/search/?searchtype=author&source=header&query=")]
         ),
--- a/plugins/src/web/config.ron
+++ b/plugins/src/web/config.ron
@@ -87,10 +87,10 @@
             matches: ["lib"],
             queries: [(name: "Libraries.io", query: "libraries.io/search?q=")]
         ),
-	(
-	    matches: ["mdn"],
-	    queries: [(name: "Mozilla Developer Network", query: "developer.mozilla.org/search?q=")]
-	),
+        (
+            matches: ["mdn"],
+            queries: [(name: "Mozilla Developer Network", query: "developer.mozilla.org/search?q=")]
+        ),
         (
             matches: ["npm"],
             queries: [(name: "npm", query: "npmjs.com/search?q=")]
@@ -115,10 +115,10 @@
             matches: ["stack"],
             queries: [(name: "Stack Overflow", query: "stackoverflow.com/search?q=")]
         ),
-	(
-	    matches: ["sp", "startpage"],
-	    queries: [(name: "Startpage", query: "startpage.com/sp/search?query=")]
-	),
+        (
+            matches: ["sp", "startpage"],
+            queries: [(name: "Startpage", query: "startpage.com/sp/search?query=")]
+        ),
         (
             matches: ["twitch"],
             queries: [(name: "Twitch", query: "twitch.tv/search?term=")]
@@ -133,7 +133,11 @@
         ),
         (
             matches: ["www"],
-            queries: [(name: "Open Website", query: "http://")]
+            queries: [(
+                name: "Open Website",
+                query: "http://",
+                icon: "https://upload.wikimedia.org/wikipedia/commons/thumb/7/72/Crystal128-browser.svg/179px-Crystal128-browser.svg.png",
+            )]
         ),
         (
             matches: ["xaut", "arxauth", "arxivauthor"],
--- a/plugins/src/web/config.ron
+++ b/plugins/src/web/config.ron
@@ -135,7 +135,7 @@
             matches: ["www"],
             queries: [(
                 name: "Open Website",
-                query: "http://",
+                query: "https://",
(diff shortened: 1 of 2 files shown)


n3m0-22 (changes requested):
I'm seeing some issues with this. No matter what browser is set to default (e.g., midori, firefox, chrome, brave, chromium) it only seems to launch Firefox. When Firefox launches it does not go to the page requested. It is also opening a new window instead of a new tab when Firefox is already open. After testing with Firefox uninstalled this launches no browser.

canadaduane:
> No matter what browser is set to default (e.g., midori, firefox, chrome, brave, chromium) it only seems to launch Firefox.

Oh interesting! I think I saw this at first as well, but didn't understand what was going on--so I then logged out and back, and it worked for me. It now launches correctly in Chromium (my system browser).

Are you seeing anything similar? Or does it only try to launch Firefox, even after logout?

canadaduane:
Also, the icon is a bit strange--but perhaps this is normal for a URL that it does not know?

image

n3m0-22:
I have tested with Logout, Reboot, Poweroff/Restart, `Alt`+ `F2` and `r`, and nothing seems to get it working on my end. As for the icon this PR https://github.com/pop-os/launcher/pull/129 may be useful. To be sure I tested on a few other machines one of which is a clean install, but still no luck.

n3m0-22:
The first machine I tested on had #155 installed and I updated from there. I didn't seem to make any difference. I did try with and without #155 JIC.

canadaduane:
Ok, I think I've figured out what's going on. There is a "panic" that causes the launcher to crash when attempting to retrieve an icon URL without a domain (i.e. `http://` with no domain).

```
Jan 12 19:44:02 rosie gnome-shell[184724]: favicon_path: /home/duane/.cache/pop-launcher/Open Website.ico
Jan 12 19:44:02 rosie gnome-shell[184724]: thread 'main' panicked at 'invalid url: EmptyHost', /home/duane/tmp/launcher/plugins/src/web/mod.rs:141:36
Jan 12 19:44:02 rosie gnome-shell[184724]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Somehow, I was able to get it to fetch some image (I'm not sure how!) and it cached it in `~/.cache/pop-launcher/Open Website.ico`. From then on, the `www` pattern to open a website worked correctly, albeit with a garbled icon image.

n3m0-22:
I went back to square one with this. I setup a fresh install then built and installed #155 and this PR. After that it worked. I then tested on another machine with the staging repo to get #155 and then added this PR. That also worked. To confirm I tested on a old VM and did the same updates. That worked as well. I'm not sure why it was not working before, but I can confidently say this is working now on my end.

@canadaduane Before I approve do you want to push the changes with the icon?

canadaduane:
Thanks for testing this again. After discovering #157 I think we are hitting a "random chance" bug that is affecting our success or failure here. It could have just been random chance that it worked for you, because it would depend on whether or not an image had enough time to cache asynchronously before panicking. So I would recommend waiting for that PR merge first.

canadaduane:
> @canadaduane Before I approve do you want to push the changes with the icon?

I tried to get an icon working, but failed :S I could just be doing it wrong. Do we have an example `config.ron` somewhere with the icon in use?

n3m0-22:
This example from #129 is working for me.

```
(
   matches: ["np"],
   queries: [(
      name: "Nix Packages", 
      query: "search.nixos.org/packages?channel=unstable&query=",
      icon: "https://nixos.wiki/nixos-logo-small.png",
   )]
),
```
If you are talking about using a local icon like something in `/usr/share/icons` I have not tested that as it was not the goal of the PR, but I can look into it.

As for #157 I'll start checking it, and keep an eye out to test any changes as they come in.

mmstick:
@canadaduane  Could you include the commits from #157 in this PR? This PR relies on that PR so it'd make sense to have them in a single PR so QA only has to test one PR.

canadaduane:
PRs merged into this one for easier QA. Thanks for the suggestion!

canadaduane:
> "https://nixos.wiki/nixos-logo-small.png"

I don't understand why, but I can't seem to get an icon working alongside this change. It shows the magnifying glass icon, which is fine. There is no accompanying "URL" icon, however. Adding the "icon:" field as @n3m0-22 suggested does not seem to work for now (FWIW I tried "https://upload.wikimedia.org/wikipedia/commons/thumb/7/72/Crystal128-browser.svg/179px-Crystal128-browser.svg.png" as an icon, but it didn't show up when `www` pattern was given in the launcher).

canadaduane:
> I don't understand why, but I can't seem to get an icon working alongside this change.

Ok, it's my PR's fault. The early return that guards against invalid pattern URLs simultaneously prevents the `icon` URL from being fetched in the backround. Working on it.

canadaduane:
Ok, I think this is ready for review again.

plugins/src/web/config.ron
```diff
@@ -131,6 +131,14 @@
             matches: ["wiki"],
             queries: [(name: "Wikipedia", query: "wikipedia.org/w/index.php?search=")]
         ),
+        (
+            matches: ["www"],
+            queries: [(
+                name: "Open Website",
+                query: "http://",
```
mmstick, line 138: Perhaps we should default to https?

canadaduane: I was really split on this. I could go either way. I think there are a very small handful of URLs I visit that are not https, but it doesn't feel right that they should make all the other URLs insecure by default.

mmstick (approved):

n3m0-22 (approved):
This all seems to be working now. The icon is working correctly. There are no issues when changing the default browser. Works with `.deb`, flatpak, and snap browsers. Looks good to me.

PR #157, opened 2023-01-13
fix(web): Ignore invalid URLs rather than panicking

canadaduane:
This fixes an issue where the launcher panics if given an invalid URL. Precondition to PR #156 working as expected.

Note that this fix prevents another (possibly bad) issue: if the launcher panics while fetching an icon image to the cache, a corrupt image may be stored. Subsequent uses of pop-launcher could attempt to display the corrupted (cached) image, resulting in either a garbled image or, perhaps, undefined behavior.

### DIFF
--- a/plugins/src/web/mod.rs
+++ b/plugins/src/web/mod.rs
@@ -131,12 +131,9 @@ impl App {
         let client = self.client.clone();
 
         let query = build_query(def, "");
-        let url = match Url::parse(&query) {
-            Ok(parsed) => parsed,
-            Err(_) => {
-                // return early if passed an invalid url, e.g. 'http://'
-                return;
-            }
+        let Ok(url) = Url::parse(&query) else {
+            // return early if passed an invalid url, e.g. 'http://'
+            return;
         };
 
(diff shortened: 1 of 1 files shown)


plugins/src/web/mod.rs
```diff
@@ -130,14 +130,21 @@ impl App {
...
-        let url = build_query(def, "");
-        let url = Url::parse(&url).expect("invalid url");
+        let query = build_query(def, "");
+        let url = match Url::parse(&query) {
+            Ok(parsed) => parsed,
+            Err(_) => {
+                // return early if passed an invalid url, e.g. 'http://'
+                return;
+            }
+        };
```
mmstick, line 140: This might be better

```rs
let Ok(parsed) = Url::parse(&query) else {
    return;
};
```

canadaduane:
Thanks! Much better.

Jan 13, 2023 7:10:56 AM Michael Murphy ***@***.***>:

mmstick (approved):

canadaduane:
This PR has been superceded by #156, as it includes this PR's changes, among others.

PR #159, opened 2023-01-15
feat(pop_shell): add config with window name/description search scope

### LINKED ISSUE #152: Match on application names rather than window names in Launcher
### Feature summary

By default, pressing Super in gnome-shell with Cosmic enabled produces a modal dialog that lets the user:
1. Switch to an open window, by name
2. Open an application, by name

However, for (1), the text is matched on the basis of the window title, not the application name. For example, if there is a web search for "emacs" open in Firefox, then typing in "emacs" may switch to Firefox, even if Emacs is open.

In my personal workflow, I would prefer it if the Launcher matched candidate windows based only (or at least first) on the name of the containing application (which is also displayed, underneath the window name).

### How would you like it to work

I imagine that users rely on the current behavior. Therefore, one way to improve Cosmic might be to add an option for what candidates should be displayed in the Launcher. I have looked through Settings and so far I haven't been able to find any options whatsoever for Launcher behavior, although I could have missed them. By contrast, similar software for other platforms (Spotlight or Alfred on macOS) has many options regarding what can be selected from the interface, and in which order candidates are displayed.

I previously filed https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/6258 because I did not realize that the behavior of gnome-shell had been changed by Cosmic. After being redirected here, I did a search of existing issues on this tracker, but did not find anything on the topic.

I am happy to prepare a pull request that either changes this behavior by default, or adds an option to customize it, if such a contribution would be acceptable.

### Relevant links, screenshots, screencasts etc.

Example of where typing in "firefox" does not select Firefox by default, but rather an Emacs window that just happens to have a file called "firefox" open (I apologize for taking a photograph; it does not seem possible to use the screenshot facility while Launcher is open):

canadaduane:
This adds an optional `config.ron` file to the `pop_shell` plugin and allows users to configure what desktop window attributes they want searched (e.g. window title and/or window description).

For example, I prefer the window name not to be searched, because it's often a web page that has set the title of the window (e.g. in Chromium). Because window items in the launcher list always sort above desktop apps, these active window HTML pages can get in the way of fast/memory muscle launcher usage.

My config is as follows (`~/.local/share/pop-launcher/plugins/pop_shell/config.ron`):

```ron
(
    search: (name: false, description: true),
)
```

With the above config, even if I'm on a google search page for `xournal++`, typing `xou` into pop-launcher will always load my desktop app, "Xournal++", even if a Chromium HTML page has set the title to "Xournal++ Search Results". (Typing "chro", however, would still switch to Chromium).

The default remains as it was, effectively:

```ron
(
    search: (name: true, description: true),
)
```

Fixes #152

### DIFF
--- a/plugins/src/pop_shell/config.rs
+++ b/plugins/src/pop_shell/config.rs
@@ -1,0 +1,55 @@
+// SPDX-License-Identifier: GPL-3.0-only
+// Copyright © 2021 System76
+
+use serde::Deserialize;
+
+pub fn bool_true() -> bool {
+    true
+}
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct SearchScope {
+    #[serde(default = "bool_true")]
+    pub name: bool,
+
+    #[serde(default = "bool_true")]
+    pub description: bool,
+}
+
+impl Default for SearchScope {
+    fn default() -> SearchScope {
+        SearchScope {
+            name: true,
+            description: true,
+        }
+    }
+}
+
+#[derive(Debug, Default, Deserialize, Clone)]
+pub struct Config {
+    #[serde(default)]
+    pub search: SearchScope,
+}
+
+pub fn load() -> Config {
+    let mut config = Config::default();
+
+    for path in pop_launcher::config::find("pop_shell") {
+        let string = match std::fs::read_to_string(&path) {
+            Ok(string) => string,
+            Err(why) => {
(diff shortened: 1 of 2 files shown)


n3m0-22 (approved):
This looks good to me. It's working as expected, and I did not find any issues.

PR #160, opened 2023-01-15
Handle shebang more thoroughly in scripts plugin

### LINKED ISSUE #142: Allow scripts to define their interpreter
Howdie, thanks so much for sharing this project <3

My scripts can be found via a launcher front-end just fine (I'm using [`onagre`](https://github.com/oknozor/onagre))

However, if there is an error in a script or other useful output, there doesn't appear to be any feedback or log entries (I've checked `journalctl -xb`, I've checked the stderr/stdout from both `onagre` and `pop-launcher`)

I noticed that we don't scripts are run in a way that throws away their stderr/stdout content: https://github.com/pop-os/launcher/blob/master/plugins/src/scripts/mod.rs#L62-L64

I understand that the API for pop-launcher and its plugins is built around JSON lines in stdin/stdout, so it would break the API to allow script execution to write to stdin/stdout, but should it be okay to pipe through stderr?

Should there be a way for pop-launcher to signal to the front-end that there was an error with an activation?

Cheers!

canadaduane:
This PR allows the shebang line in a script file to have more flexibility:

1. Whitespace is accepted after the `#!` characters, so `#! /bin/bash` is now ok.
2. Any args passed to the command in the shebang are honored, so `#!/usr/bin/env nu` will work (for example)

Fixes #142 (except for stderr and non-`#` metadata comment lines)

### DIFF
--- a/plugins/src/scripts/mod.rs
+++ b/plugins/src/scripts/mod.rs
@@ -54,12 +54,27 @@ impl App {
     }
 
     async fn activate(&mut self, id: u32) {
+        let whitespace_re = Regex::new(r"[ \t]+").unwrap();
+
         if let Some(script) = self.scripts.get(id as usize) {
             let interpreter = script.interpreter.as_deref().unwrap_or("sh");
+
+            // split the shebang into parts, e.g. ["/bin/bash"], or a more complex ["/usr/bin/env", "bash"]
+            let parts: Vec<&str> = whitespace_re.split(interpreter).collect();
+
+            // first token must be the command to run, e.g. "/usr/bin/env"
+            let program: &str = parts.first().unwrap();
+
+            // keep args as given, e.g. "bash"
+            let mut args: Vec<&str> = parts[1..].to_vec();
+
+            // finally, add the script file itself as a final arg
+            args.push(script.path.to_str().unwrap());
+
             send(&mut self.out, PluginResponse::Close).await;
 
-            let _ = Command::new(interpreter)
-                .arg(script.path.as_os_str())
+            let _ = Command::new(program)
+                .args(args)
                 .stdin(Stdio::null())
                 .stdout(Stdio::null())
                 .stderr(Stdio::null())
--- a/plugins/src/scripts/mod.rs
+++ b/plugins/src/scripts/mod.rs
@@ -6,6 +6,7 @@ use pop_launcher::*;
 
 use flume::Sender;
 use futures::StreamExt;
+use regex::Regex;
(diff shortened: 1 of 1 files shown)


plugins/src/scripts/mod.rs
```diff
@@ -53,12 +54,27 @@ impl App {
...
+            let mut args: Vec<&str> = parts[1..].to_vec();
+
+            // finally, add the script file itself as a final arg
+            args.push(script.path.to_str().unwrap());
+
             send(&mut self.out, PluginResponse::Close).await;
 
-            let _ = Command::new(interpreter)
-                .arg(script.path.as_os_str())
+            let _ = Command::new(program)
```
jokeyrhyme, line 76: Huh, is there a reason why we don't use `Command::new(script.path)` instead and rely on the OS/shell/binfmt/whatever to execute the script? That way we don't necessarily even need to parse the sha-bang line at all

canadaduane: I'm fairly new to Rust, but I read the `Command::new` docs as saying the string passed in must be either an abs path to the executable, or the command to search for on the path, with no args:

```
Constructs a new Command for launching the program at path program, with the following default configuration:

No arguments to the program
...
```

https://doc.rust-lang.org/std/process/struct.Command.html

canadaduane: Oh, I see what you're saying. Just pass the script itself in. Maybe that would work!

jokeyrhyme: Maybe, but there's probably a reason why the original code worked the way it did, and it old code does have the benefit of only needing read-permissions on the script file :shrug:

plugins/src/scripts/mod.rs
```diff
@@ -53,12 +54,27 @@ impl App {
...
+            let parts: Vec<&str> = whitespace_re.split(interpreter).collect();
+
+            // first token must be the command to run, e.g. "/usr/bin/env"
+            let program: &str = parts.first().unwrap();
+
+            // keep args as given, e.g. "bash"
+            let mut args: Vec<&str> = parts[1..].to_vec();
+
+            // finally, add the script file itself as a final arg
+            args.push(script.path.to_str().unwrap());
```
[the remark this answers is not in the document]
canadaduane: Thanks! Glad to learn here. PR updated.

mmstick (approved):

n3m0-22 (approved):
This looks good to me. I have tested the issues listed and they seem to be fixed by this. I see no regressions.

PR #161, opened 2023-01-15
feat(find): let fdfind use --full-path search to include directories

### LINKED ISSUE #150: support directory search
I like launcher very much. Most of my works need。
But I have many front-end projects. you now,  with thousands of node_modules packages.
They confuse me when I use launcher to find my project, (for example, I want to enter one of projects, but I cannot search my project name)
So I need it very much.

canadaduane:
Allows the user to find files within directories. Useful when files with the same name are found in many directories, e.g. `package.json`:

```
find MyProject/package.json
```

Fixes #150


mmstick (approved):

n3m0-22 (approved):
This is working as expected. I can now search for files and directories. I'm also now seeing files with the same names and the directories they are each in. Looks good to me. Thank you for the contribution.

Resolves #150

Issue #165, opened 2023-01-26
Feature Request: Lock screen option


btegs:
If I open up the Launcher and type any of these, they will work:

* Log Out
* Power Off/Shutdown
* Restart
* Suspend

But if I type **Lock**, I do not see the option to Lock the screen. Is this something planned or easy to implement? Just a little quality of life thingie :-)

n3m0-22:
I've created a PR #166 that will need to be reviewed and tested. In the meantime you can lock the screen with `Super` + `Esc`.

You can also make your own script for the time being with the following commands.

- `touch ~/.local/share/pop-launcher/scripts/session/lock-screen.sh`
- `gedit ~/.local/share/pop-launcher/scripts/session/lock-screen.sh`

- In the file that opens paste the following:

```
#!/usr/bin/sh
#
# name: Lock
# icon: locked
# description: Lock the system
# keywords: lock
 
set -eu
 
loginctl lock-session
```
- Save the file

- `sudo chmod a+x ~/.local/share/pop-launcher/scripts/session/lock-screen.sh`

leviport:
Closing since https://github.com/pop-os/launcher/pull/166 was merged. It should be released some time this week.

btegs:
Thanks for your contributions everyone!

PR #166, opened 2023-01-26
add screen lock script

n3m0-22:
Based on the feature request https://github.com/pop-os/launcher/issues/165


jacobgkau (commented):
The new script shows up in search results and works to lock the screen.

Screenshot from 2023-01-27 12-04-35

Is there a reason the description is `Lock the system` rather than `Lock the screen`? The latter sounds more specific to me, and the entry that GNOME puts in its Applications menu is called `Lock Screen`. The person who opened the linked feature request also referred to it this way.

Edit: I see `Power off` and `Reboot` both say `the system` in their descriptions. But `Log Out` says `Log out to the login screen`.

n3m0-22:
> Is there a reason the description is `Lock the system` rather than `Lock the screen`?
No it was just a mistake on my part. Thanks for catching that.

jacobgkau (approved):
Looks great. Still works.

Screenshot from 2023-01-27 13-06-49

mmstick (approved):

PR #168, opened 2023-01-30
use xdg-screensaver in session-lock script for cross compatibility

n3m0-22:
This change is based on comments from the original PR #166.

https://github.com/pop-os/launcher/commit/970e71e2beec1a381e879e5df6652b47f6a017e0#commitcomment-98615497

https://github.com/pop-os/launcher/commit/970e71e2beec1a381e879e5df6652b47f6a017e0#commitcomment-98616065

In my initial testing with using `xdg-screensaver` I saw it fail a few times on the `lemp11`. This my have been a one off, but should be tested for.

### DIFF
--- a/scripts/session/session-lock.sh
+++ b/scripts/session/session-lock.sh
@@ -7,4 +7,4 @@
 
 set -eu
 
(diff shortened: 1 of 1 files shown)


mmstick (approved):

leviport (approved):
Works as expected in Gnome

PR #172, opened 2023-02-18
[WIP] Flexible search plugin (via Unix pipes)

### LINKED ISSUE #164: Proposal: Make the `find` built-in more versatile
I recently [modified the `find` plugin](https://github.com/pop-os/launcher/pull/161) to add the `--full-path` arg to the underlying `fdfind` CLI.

In doing so, it seemed to me there were really a lot of options that could be quite reasonable for `fdfind`, notably the `--glob` parameter, which makes the launcher's `find` command behave more like a command-line pattern match (as opposed to the `regex` style pattern match that is currently the default).

My first impulse was to make the `find` command take a `config.ron` that allows to choose `glob` vs `regex`. However, I realized the entire `find` plugin is a great example of a potentially reusable pattern--

1. Given a CLI command that lists results as output (e.g. `fdfind`)
2. and a set of arguments to pass in as defaults (e.g. `--full-path`, `--regex`, `--glob`, etc.)
3. pass the search-term as last argument to the CLI tool
4. display each line from STDOUT as a corresponding result in the launcher
5. pass the result to a command to run when executed

To my surprise, this pattern is actually already built for URLs--it's the `web` plugin! We just need something that behaves this way for CLI tools.

Would it make sense to either

- (a) modify the `find` plugin to be more versatile so that it has a `config.ron` similar to the `web` plugin, but to use CLI commands instead of `xdg-open`, or
- (b) create an entirely new plugin that does the same?

I'm kind of thinking (a) makes more sense since if I build this, then the current `find` command would become simply a setting to add to the `config.ron` of this new plugin (i.e. the new plugin would functionally be a superset of the `find` plugin).

Any thoughts or guidance before I dig in?

canadaduane:
NOTE: This is a WORK IN PROGRESS, and is not yet ready to be merged. Looking for early feedback, as well as rust tips and improvements to memory management and/or style. Also looking for discussion around which existing launcher plugins this could/should replace, if any.

## Intro

This is a generalized "search" plugin that can be configured to pass a query from the Pop Launcher frontend to any shell command that returns results (the shell command must return one result per line).

For example, the current "find" plugin (which has 245 lines of rust code), can be implemented in about 10 lines of config, and offers more flexibility to the end user since it becomes trivial to modify the `fdfind` shell args if desired:

```ron
# ~/.local/share/pop-launcher/plugins/search/config.ron
(
  rules: [
    (
      pattern: StartsWithKeyword(["f", "find"]),
      action: (
        query_command: "fdfind --ignore-case --full-path $KEYWORD1",
        output_captures: "^(.+)/([^/]+)$",
        result_name: "$CAPTURE2",
        result_desc: "$CAPTURE1",
        run_command: "xdg-open '$OUTPUT'",
      )
    ),
  ]
)
```

The `pattern` line indicates when to match the rule. It corresponds with the "prefix" that you type into the launcher frontend. In the above example, the search plugin would begin an `fdfind` query if the user starts typing "f" or "find" into the launcher frontend.

The `result_name` and `result_desc` settings correspond to the 1st and 2nd lines of each search result (the launcher frontend can show two lines per result). The `result_name` and `result_desc` settings are optional. By default `result_name` is "$OUTPUT", and `result_desc` is blank. Keywords from the query string, and captures from the `output_captures` regex can be used as needed to format a search result for the user.

The `query_command` is the shell command to use as the query function, e.g. `fdfind` in our example above. As long as the shell command returns only one line (on STDOUT) per result, any command may be used. For example, `ls -1`, `apt list`, a [shell command that lists your address book contacts](https://github.com/francoisfreitag/mutt-gnome-contacts), etc.

Finally, the `run_command` is the command to run on the active result chosen by the user, i.e. what to do when the user presses "Enter" in the launcher frontend.

Note that variables are expanded using [shellexpand](https://docs.rs/shellexpand/latest/shellexpand/fn.env_with_context.html) and therefore can use curly braces and default values, e.g. `${KEYWORD1:-someDefault}`.

### Example 1: List hard drives

A launcher command, "drives" that shows a list of block devices:

```ron
    (
      pattern: StartsWithKeyword(["drives"]),
      action: (
        query_command: "lsblk -lno NAME,SIZE,MOUNTPOINTS",
        run_command: "notify-send '$OUTPUT'",
      )
    ),
```

### Example 2: Show running processes

A launcher command, "ps" that lists processes, their PIDs, and how much CPU they are using. This example demonstrates using bash to write a "mini shell script" with a pipe between `ps` and `head`:

```ron
    (
      pattern: StartsWithKeyword(["ps"]),
      action: (
        query_command: "/bin/bash -c 'ps --sort=-pcpu -axo pid,ucmd,pcpu | head -10'",
        output_captures: "^\\s+([0-9]+)\\s+(.*)$",
        result_name: "${CAPTURE2}",
        result_desc: "${CAPTURE1}",
        run_command: "notify-send '$OUTPUT'",
      )
    ),
```

### Example 3: Search apt packages

A launcher command, "apt" that searches for packages using glob syntax of `apt list`, e.g. "apt zu*":
```ron
    (
      pattern: StartsWithKeyword(["apt"]),
      action: (
        query_command: "apt list $KEYWORD1",
        output_captures: "^([^/]+)/(.+)$",
        result_name: "$CAPTURE1",
        result_desc: "$CAPTURE2",
        run_command: "notify-send '$OUTPUT'",
      )
    ),
```

To understand the above, it's helpful to know how the shell command `apt list [search_term]` sends results to STDOUT--one per line, in the following format:

```
$ apt list zu*
Listing... Done
zulucrypt-cli/jammy 5.7.1-2 amd64
zulucrypt-gui/jammy 5.7.1-2 amd64
zulumount-cli/jammy 5.7.1-2 amd64
zulumount-gui/jammy 5.7.1-2 amd64
zulupolkit/jammy 5.7.1-2 amd64
zulusafe-cli/jammy 5.7.1-2 amd64
zurl/jammy 1.11.1-1 amd64
zutils/jammy 1.11-1 amd64
zutty/jammy 0.11.2.20220109.192032+dfsg1-1 amd64
```
Note that "Listing... Done" is skipped, because it does not match the `output_capture`'s regular expression, which requires that a "/" must be present in the search result line.

### Example 4: Calculator

```ron
    (
      pattern: StartsWith(["="]),
      split: Regex("^="),
      action: (
        query_command: "qalc -u8 -set 'maxdeci 9' -t $KEYWORD1",
        result_name: "$KEYWORD1",
        result_desc: "$OUTPUT",
        run_command: "/bin/bash -c 'wl-copy \"$OUTPUT\" && notify-send \"Copied to clipboard\"'",
      )
    ),
```

The `split` setting is new in the example above: it allows us to split the query using a regular expression. Normally, the query is split into keywords using the default "ShellWords" algorithm, which splits on spaces but also understands things like single and double quotes (treating words enclosed in quotes as a single entity). Here, however, to enter calculator mode with an initial "=" regardless of whether a space follows the "=", we need to split the query differently. In this example, we choose to split on the initial "=", making `$KEYWORD1` absorb the remaining equation to be sent to `qalc` as input (as well as the frontend as the name).

Fixes #164

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1534,6 +1534,8 @@ dependencies = [
  "ron",
  "serde",
  "serde_json",
+ "shell-words",
+ "shellexpand",
  "slab",
  "strsim",
  "sysfs-class",
@@ -1969,6 +1971,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "900fba806f70c630b0a382d0d825e17a0f19fcd059a2ade1ff237bcddf446b31"
 dependencies = [
  "lazy_static",
+]
+
+[[package]]
+name = "shell-words"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "24188a676b6ae68c3b2cb3a01be17fbf7240ce009799bb56d5b1409051e78fde"
+
+[[package]]
+name = "shellexpand"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dd1c7ddea665294d484c39fd0c0d2b7e35bbfe10035c5fe1854741a57f6880e1"
+dependencies = [
+ "dirs 4.0.0",
 ]
 
 [[package]]
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -20,8 +20,8 @@ async fn main() {
         match cmd {
             "calc" => plugins::calc::main().await,
             "desktop-entries" => plugins::desktop_entries::main().await,
-            "find" => plugins::find::main().await,
             "files" => plugins::files::main().await,
+            "search" => plugins::search::main().await,
             "pop-launcher" => service::main().await,
             "pop-shell" => plugins::pop_shell::main().await,
             "pulse" => plugins::pulse::main().await,
--- a/plugins/Cargo.toml
+++ b/plugins/Cargo.toml
@@ -33,6 +33,8 @@ dirs = "4.0.0"
 futures = "0.3.25"
 bytes = "1.2.1"
 recently-used-xbel = "1.0.0"
+shell-words = "1.1.0"
+shellexpand = "3.0.0"
 
 # dependencies cosmic toplevel
 cctk = { git = "https://github.com/pop-os/cosmic-protocols", package = "cosmic-client-toolkit" }
--- a/plugins/src/lib.rs
+++ b/plugins/src/lib.rs
@@ -6,6 +6,7 @@ pub mod cosmic_toplevel;
 pub mod desktop_entries;
 pub mod files;
 pub mod find;
+pub mod search;
 pub mod pop_shell;
 pub mod pulse;
 pub mod recent;
--- a/plugins/src/search/app.rs
+++ b/plugins/src/search/app.rs
@@ -9,7 +9,7 @@ use tokio::process::ChildStdout;
 use pop_launcher::{async_stdout, PluginResponse, PluginSearchResult};
 
 use crate::search::config::Definition;
-use crate::search::util::{interpolate_result, interpolate_run_command};
+use crate::search::util::{interpolate_result, interpolate_run_command, split_query};
 
 use super::config::{load, Config};
 use super::util::{exec, interpolate_query_command};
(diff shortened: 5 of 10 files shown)


mmstick:
Some things that would improve this:

- Each command should have its own configuration file somewhere, similar to how scripts are in `/usr/lib/pop-launcher/scripts`
- Some way of enabling and disabling a command, perhaps with command line arguments to `pop-launcher` itself
- Ability to define a condition(s) that must be true for the command to be queried; such as `Exists("apt")`.
- Custom icons for each command, and custom icons for each result

I'm also wondering if we could merge this and the scripts plugin together with a solution based on [rhai](https://crates.io/crates/rhai). It's possible to pass values and functions from the Rust side to the Rhai scripts, so there could be a function for setting up Rhai-based plugins in this way.

canadaduane:
Thanks for the feedback! I have a few questions:

> Each command should have its own configuration file somewhere, similar to how scripts are in /usr/lib/pop-launcher/scripts

I'm certainly willing to do this, but I'm not sure if I understand the benefit. Is this to enable programs/packages to install their own launcher scripts without having to mutate the "big config file"?

> Some way of enabling and disabling a command, perhaps with command line arguments to pop-launcher itself

I'm guessing this is in anticipation of a settings/configuration page of some kind, so we can let the user decide what system-installed scripts are enabled/disabled? Should we require a "search plugin" config script to have an ID? This would allow a user to override the system script with their own, while still allowing a settings page to enable/disable.

> Ability to define a condition(s) that must be true for the command to be queried; such as Exists("apt").

Cool, good idea.

> Custom icons for each command, and custom icons for each result

Yes, I was thinking the same.

> I'm also wondering if we could merge this and the scripts plugin together with a solution based on [rhai](https://crates.io/crates/rhai). It's possible to pass values and functions from the Rust side to the Rhai scripts, so there could be a function for setting up Rhai-based plugins in this way.

I'm a bit intimidated by the complexity of this, in conjunction with the focus that this search plugin has on "just piping". Would it be OK to create a 2nd plugin with Rhai as the focus? Or is there a distinct advantage to combining that capability with this search plugin's capability?

canadaduane:
@mmstick If you have a moment, I'm looking for feedback before proceeding.

mmstick:
Sorry for the delay.

I want to simplify configuration in the future. If each plugin has its own configuration file, it will be easier for people to install third party configurations without having to append them.

I may also make the transition from RON to KDL for first party plugins since that's easier for a human for a human to edit. Much reduced chance of syntax errors.

There may eventually be a settings page for the launcher and its plugins, and it's a common request to disable a plugin or an item from a plugin.

I can convert plugins to Rhai in the future. Same with KDL.

AnomalRoil:
I came here looking for this initially. I'm missing `rofi -dmenu` / `dmenu` since I started running Cosmic and would have liked a way to re-use the existing `cosmic-launcher` for the prompting, without having to change my existing scripts too much. Sadly couldn't find a way.

I ended up implementing my own plugin in `~/.local/share/pop-launcher/plugins/` and it was relatively easy, except for the fact that the IPC communication isn't super well documented and I ended up initially not sending enough `"Finished"` on my `stdout`. 

But yeah, not sure if it's best to have a "dmenu plugin" mode for the launcher or a `dmenu` flag / command for `cosmic-launcher.

PR #175, opened 2023-03-14
fix(cosmic_toplevel plugin): add window to search result response

wash2:


Drakulix (approved):

jacobgkau (approved):
@wash2 Since you're an employee, please put your branches directly on the repo (in the pop-os org) and don't fork or use colons in branch names. Forking keeps Jenkins from building the branch and makes it less convenient for QA to test.

<details>
<summary>Regression testing passed in GNOME (X11):</summary>

</details>

I'm not sure what this is needed for in COSMIC, but the launcher seems to work both without and with it. (Not sure if this goes with https://github.com/pop-os/cosmic-launcher/pull/33? I didn't notice a difference applying that one either.)

wash2:
> I'm not sure what this is needed for in COSMIC, but the launcher seems to work both without and with it. (Not sure if this goes with pop-os/cosmic-launcher#33? I didn't notice a difference applying that one either.)

Ya it's for cosmic-launcher. I haven't merged the PR in cosmic-launcher that uses this yet though

PR #176, opened 2023-03-16
Add session script for rebooting into BIOS

ahoneybun:
This is copied from the session-reboot.sh script but changed to reboot into BIOS, this can be handy for quickly making changes and not needing to use the BIOS key during boot.

### DIFF
--- a/scripts/session/session-bios.sh
+++ b/scripts/session/session-bios.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/sh
+#
+# name: Enter BIOS
+# icon: system-restart
+# description: Reboot into BIOS
+# keywords: bios uefi reboot restart 
+
+set -eu
+
+is_gnome() {
+  command -v dbus-send >/dev/null && \
+    command -v gnome-session-quit >/dev/null && \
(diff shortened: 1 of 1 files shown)


scripts/session/session-bios.sh
```diff
@@ -0,0 +1,20 @@
...
+set -eu
+
+is_gnome() {
+  command -v dbus-send >/dev/null && \
+    command -v gnome-session-quit >/dev/null && \
+    dbus-send --print-reply --dest=org.gnome.Shell /org/gnome/Shell org.freedesktop.DBus.Properties.Get string:org.gnome.Shell string:ShellVersion >/dev/null 2>&1
+}
+
+if is_gnome; then
+  gnome-session-quit --reboot
```
mmstick, line 17: So GNOME sessions will not reboot into firmware setup?

ahoneybun: Should be fixed now, this works for me.

n3m0-22 (approved):
This looks good to me. I've tested on several systems with coreboot as well as proprietary firmware. It has worked on all tests.

mmstick (approved):

PR #183, opened 2023-06-03
fix(service): be consistent in getting cache path

rmsthebest:
very minor fix, but it was a bit inconsistent in following xdg dirs.
I also slightly changed wording in the error messages as they were a bit confusing to me when I was seeing them.

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -5,6 +5,7 @@ use pop_launcher_toolkit::plugins;
 use pop_launcher_toolkit::service;
 
 use mimalloc::MiMalloc;
+use pop_launcher_toolkit::service::ensure_cache_path;
 
 #[global_allocator]
 static GLOBAL: MiMalloc = MiMalloc;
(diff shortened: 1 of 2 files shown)


nkia-christoph (approved):

jacobgkau (commented):
This causes the `recent` file to be created in `~/.local/state/pop-launcher/` instead of `~/.cache/pop-launcher/`. It still seems to be functioning as expected. Reviewing documentation and discussion about the difference between cache and state in the XDG spec, I think state is the right place for it to go.

This will effectively reset everyone's `recent` state when they install this update since the file is created from scratch in the new location (with the old one continuing to sit around in perpetuity). I don't think this is a critical issue, but it would be nice if there was some way to not have to do that... e.g. have some code to move the file to the new location, which we could remove after a few months or something.

Thoughts from anyone else if that's feasible or worth the hassle?

jacobgkau:
If we're going to make this change (without migrating the file), perhaps the 24.04 upgrade would be a good time to do it.

To ask again, would transition code to check for the old location and copy it to the new one be feasible?

mmstick:
It should move the file automatically when found

PR #195, opened 2023-09-29
cosmic-toplevel: Also match by StartupWMClass

Drakulix:
The first commit changes the `cosmic-toplevel` plugin to always parse the desktop entries and match by the parsed `app_id` (which can be different than the file name) and parsed `StartupWMClass`.

The latter fixes icons not showing up primarily for Xwayland apps, who use the wm-class for identification, but don't always use the app-id in place. Additionally this fixes a lot of electron apps, that don't handle the app-id correctly in general up until the very latest release.

The second commit introduces the same fallback icon we are using in the `cosmic-toplevel` and `pop-shell` plugins for apps lacking an icon to the `desktop-entries` plugin as well. In case this in controversial, as this icon might now show up in places, where previously no icon was shown, this commit can easily be dropped. In my opinion is improves consistency across the plugins.

### DIFF
--- a/plugins/src/cosmic_toplevel/mod.rs
+++ b/plugins/src/cosmic_toplevel/mod.rs
@@ -16,7 +16,7 @@ use pop_launcher::{
     Request,
 };
 use std::borrow::Cow;
-use std::{ffi::OsString, fs, path::PathBuf};
+use std::{fs, path::PathBuf};
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 
 use self::toplevel_handler::{toplevel_handler, ToplevelAction, ToplevelEvent};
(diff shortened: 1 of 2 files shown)


mmstick (approved):

leviport (approved):
I'm not seeing any regressions on GNOME:

### x11
- [x] All windows on all workspaces appear on launch
- [x] Choosing an app on another workspace moves workspaces and focus to that app
- [x] Launching an application works
- [x] Typing text and then removing it will re-show those windows
- [x] Search works for applications and windows
- [x] Search works for GNOME settings panels
- [x] Search for "Extensions". There should be only one entry.
- [x] The overlay hint correctly highlights the selected window
- [x] Open windows are sorted above applications (e.g. "web browser")
- [x] t: executes a command in a terminal
- [x] : executes a command in sh
- [x] = calculates an equation
- [x] Search results are as expected:

### Wayland
- [x] All windows on all workspaces appear on launch
- [x] Choosing an app on another workspace moves workspaces and focus to that app
- [x] Launching an application works
- [x] Typing text and then removing it will re-show those windows
- [x] Search works for applications and windows
- [x] Search works for GNOME settings panels
- [x] Search for "Extensions". There should be only one entry.
- [x] The overlay hint correctly highlights the selected window
- [x] Open windows are sorted above applications (e.g. "web browser")
- [x] t: executes a command in a terminal
- [x] : executes a command in sh
- [x] = calculates an equation
- [x] Search results are as expected:

PR #205, opened 2024-01-30
desktop_entries: Allow selecting gpu/secondary actions on COSMIC

Drakulix:
Adds entries for each gpu and secondary desktop actions to the context menu of desktop entries, while keeping compatibility with pop-shell, by carefully adjusting the protocol and not exposing different gpus (which the shell-plugin can't handle).

Only change for pop-shell is the use of switcheroo to determine, if the setup should be considered "switchable".

Testing on COSMIC requires https://github.com/pop-os/cosmic-launcher/pull/85


mmstick (approved):

leviport (dismissed):
The same "Launch with Discrete GPU" prompt still appears in Gnome. The prompt cannot be clicked, but that seems to be an existing bug, not a regression. I see no other adverse behavior with the launcher on Gnome.

Drakulix:
Needs another review, due to small fix for systems with switcheroo running but no gpu (VMs with software rendering) and the "default"-label not respecting the "prefers_dgpu" property, although launching the app with the correct preference.

leviport (approved):
No change in behavior on Gnome on switchable and integrated laptops. Also works in VMs.

mmstick (approved):

PR #213, opened 2024-05-03
Toplevel high priority

wash2:
This should help fix https://github.com/pop-os/cosmic-launcher/issues/121. While, I believe they do currently show up with a specific enough search, cosmic_toplevel matches are listed at the top of the list when it is set to high priority.


mmstick (approved):

XV-02 (approved):
This looks to resolve the associated issue for the Cosmic DE launcher, and I'm not seeing any regressions with the changes for the Gnome or Cosmic implementations of our launcher. This appears to be good to go.

PR #214, opened 2024-05-14
feat: prefer cosmic-term in terminal plugin

wash2:
fixes https://github.com/pop-os/cosmic-launcher/issues/82

### DIFF
--- a/plugins/src/terminal/mod.rs
+++ b/plugins/src/terminal/mod.rs
@@ -116,7 +116,46 @@ impl App {
 }
 
 fn detect_terminal() -> (PathBuf, &'static str) {
+    use freedesktop_desktop_entry::DesktopEntry;
     use std::fs::read_link;
+
+    let terminal_apps: Vec<_> =
+        freedesktop_desktop_entry::Iter::new(freedesktop_desktop_entry::default_paths())
+            .filter_map(|path| {
+                std::fs::read_to_string(&path).ok().and_then(|input| {
+                    DesktopEntry::decode(&path, &input).ok().and_then(|de| {
+                        if de.no_display()
+                            || de
+                                .categories()
+                                .map(|c| c.split_terminator(';').all(|c| c != "TerminalEmulator"))
+                                .unwrap_or(true)
+                            || de.exec().is_none()
+                        {
+                            return None;
+                        }
+
+                        Some((de.id().to_owned(), de.exec().unwrap().to_owned()))
+                    })
+                })
+            })
+            .collect();
+
+    for id in ["com.system76.CosmicTerm"] {
+        for (terminal_id, exec) in &terminal_apps {
+            if terminal_id.as_str() == id {
+                return (PathBuf::from(exec), "-e");
+            }
+        }
+    }
+
+    if let Some((id, exec)) = terminal_apps.first() {
+        let arg = if id == "org.gnome.Terminal" {
+            "--"
+        } else {
+            "-e"
+        };
+        return (PathBuf::from(exec), arg);
+    }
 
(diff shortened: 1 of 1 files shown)


mmstick (approved):

plugins/src/terminal/mod.rs
```diff
@@ -118,6 +122,15 @@ impl App {
 fn detect_terminal() -> (PathBuf, &'static str) {
     use std::fs::read_link;
 
+    let cosmic_term = Path::new("/usr/bin/cosmic-term");
```
Drakulix, line 125: It's a little ugly to hardcode /usr/bin. This means installing cosmic-term in a local prefix, or /usr/local/bin or on nix will not trigger this behavior. That probably isn't tooooo bad, but maybe we should just bite the bullet and parse $PATH?

wash2: Ya, I was thinking so too, but the exiting method of selecting a terminal for the plugin also used /usr/bin. I can change it though.

jackpot51: There is code in cosmic-files to detect terminal apps. That may be useful.

wash2: I've adjusted it to work more like cosmic-files, but I wonder if all terminal apps will be compatible, or if I should ignore the other terminal apps found searching the desktop entries when cosmic-term isn't found. alacritty seems to only support `-e` or `--command`, for example.

mmstick (dismissed):

plugins/src/terminal/mod.rs
```diff
@@ -116,8 +116,42 @@ impl App {
...
+    for id in ["com.system76.CosmicTerm"] {
+        for (terminal_id, exec) in &terminal_apps {
+            if terminal_id.as_str() == id {
+                return (PathBuf::from(exec), "--");
+            }
+        }
+    }
+
+    if let Some((_, exec)) = terminal_apps.first() {
+        return (PathBuf::from(exec), "--");
```
jackpot51, line 152: -- is not guaranteed to work. See xdg-terminal-exec, it recommends -e unless another argument is specified.

jackpot51: https://github.com/Vladimir-csp/xdg-terminal-exec?tab=readme-ov-file#desktop-entry-for-a-terminal

plugins/src/terminal/mod.rs
```diff
@@ -116,8 +116,47 @@ impl App {
...
+    for id in ["com.system76.CosmicTerm"] {
+        for (terminal_id, exec) in &terminal_apps {
+            if terminal_id.as_str() == id {
+                return (PathBuf::from(exec), "-e");
+            }
+        }
+    }
+
+    if let Some((id, exec)) = terminal_apps.first() {
+        let arg = if id == "org.gnome.Terminal" {
```
jackpot51, line 152: gnome-terminal also supports `-e`, for now.

jackpot51: This is probably good enough for now, we might be able to push xdg-terminal-exec over the finish line.

wash2: Ya, I just saw that "-e" is deprecated for gnome-terminal, so I figured it might be best to do "--" instead

jackpot51 (approved):

XV-02:
Nothing I have tried has failed yet. Are there any specific use cases that I should go out of my way to test?

wash2:
> Nothing I have tried has failed yet. Are there any specific use cases that I should go out of my way to test?

If you haven't tried without cosmic-term installed, but some other terminal like alacritty or konsole that might be good.

PR #216, opened 2024-06-02
use free_desktop_entry to search desktop file

wiiznokes:
This depend on https://github.com/pop-os/freedesktop-desktop-entry/pull/15.

This will have the benefit of
- one api to get all desktop files
- cache them in your struct without a wrapper
- one api to search desktop files
- localisation support for searching entries

Currently, this works really great. The searching function still need refinement, like decide what fields to include in the search, give some field more weight, ect. But any improvement will benefit all cosmic component now.

Also, idk it it's possible yet, but it will be nice to give the returned response a score so that good matches get displayed first. I think this should be managed in the frontend since there can be multiple plugin at once, but i'm not sure.

### DIFF
--- a/plugins/Cargo.toml
+++ b/plugins/Cargo.toml
@@ -9,7 +9,9 @@ publish = false
 [dependencies]
 async-pidfd = "0.1.4"
 fork = "0.1.23"
-freedesktop-desktop-entry = "0.5.2"
+freedesktop-desktop-entry = { path = "../../freedesktop-desktop-entry", features = [
+    "matching",
+] }
 human_format = "1.1.0"
 human-sort = "0.2.2"
 new_mime_guess = "4.0.1"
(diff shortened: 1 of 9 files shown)


plugins/Cargo.toml
```diff
@@ -9,7 +9,9 @@ publish = false
 [dependencies]
 async-pidfd = "0.1.4"
 fork = "0.1.23"
-freedesktop-desktop-entry = "0.5.2"
```
mmstick, line 12: This can be set to 0.5.3

wiiznokes:
I think having a git dependencie for fde is fine here since it will likely require some change in the matching function while cosmic approaching its release

plugins/Cargo.toml
mmstick: This is now 0.5.4

wiiznokes: Okay done

plugins/Cargo.toml
```diff
@@ -9,7 +9,7 @@ publish = false
...
```
mmstick, line 12: Must update to 0.6.0 now

mmstick (changes requested):

plugins/Cargo.toml
wiiznokes: Done

mmstick (approved):

jacobgkau (approved):
I'm not seeing any user-facing difference in search behavior while testing this.

<details>

</details>

Issue #217, opened 2024-06-02
Rfc: add a score field to the PluginSearchResult


wiiznokes:
For the top_level and desktop entries plugins, a f64 between 0 and 1 will fit.

wiiznokes:
I just saw that the service does it. But still, plugins should be able to have an impact on the final result. It's great that the service handle recent selections and stuff tho

wiiznokes:
Ok so after looking the code a bit, the service actually make a pretty good job at sorting the SearchResult. Comparing the score directly have similar result.
But the problem is in the plugin priority. The top_level entries will always be shown before the desktop entries, even if they score its 0.1 and the other 1.
This problem is also present in the cosmic-launcher UI. The app will try to gather all plugin search result together, by sorting the Indice field. I'm not sure its a good idea, but mix different plugin results might be confusing. But the current behavior is not acceptable

jacobgkau:
TopLevel showing up first is intentional: https://github.com/pop-os/launcher/pull/213

PR #219, opened 2024-06-08
Desktop entries update

wiiznokes:
- run cargo clippy and fmt on the all project
- update edition
- update fde (depend on https://github.com/pop-os/freedesktop-desktop-entry/pull/20)
- update the desktop_entries plugin to take advantage of the matching function of fde

~~I plan to start playing with the score now, for modifying to order of the result. I don't include it in this PR to keep things separate, but since the new matching function will match more things, currently, the search result are not as good as previously.~~ see #217


mmstick (dismissed):

jacobgkau (changes requested):
I noticed a search behavior change that seems odd. Searching for "test", I used to get the Blackmagic RAW Speed Test app I have installed on this system as the first result:

Screenshot from 2024-06-26 11-42-57

On this branch, it doesn't show up as a result at all when searching for that string:

Screenshot from 2024-06-26 11-43-17

Searching for something more specific does yield the result, so it's still being detected as an entry by the launcher:

Screenshot from 2024-06-26 11-43-46

Any idea why this update would cause fuzzy matches to have priority over an exact match? Not matching an exact term anymore seems like a regression.

wiiznokes:
Yes, I think i understand why this is the case. The old impl use

```
let append = search_interest.starts_with(&*query)
                    || query
                        .split_ascii_whitespace()
                        .any(|query| search_interest.contains(&*query))
                    || strsim::jaro_winkler(&*query, &*search_interest) > 0.6;
```

for matching while the new only use `strsim::jaro_winkler(&*query, &*search_interest)` (in fde).

Having both 
```
search_interest.starts_with(&*query)
                    || query
                        .split_ascii_whitespace()
                        .any(|query| search_interest.contains(&*query))
```
seems redundant. I will add the `contains` condition and update you

jacobgkau:
No rush, but just to make sure this isn't waiting on us again, I assume something will need to be done in this PR to bring https://github.com/pop-os/freedesktop-desktop-entry/pull/23 in? Not sure if the `freedesktop-desktop-entry` version needs to be bumped, but we at least need a new git commit hash to trigger a rebuild.

wiiznokes:
I assume fde needs a new version

jacobgkau:
@mmstick Can you make a crate release including https://github.com/pop-os/freedesktop-desktop-entry/pull/23 so we can bump the version used in this PR to fix that regression?

mmstick (dismissed):

wiiznokes:
@jacobgkau Sorry for the ping, i just making sure you aware that this pr is now ready for review again :)

jacobgkau (changes requested):
On 4f281f7, the `Blackmagic RAW Speed Test` app now shows up when searching `test`. However, searching for `settings` does not return the Settings app (a.k.a. GNOME Control Center) as a result. On this machine, it currently shows up as the second option when searching for that term on the released/master branch.

Screenshot from 2024-07-09 17-33-06
Screenshot from 2024-07-09 17-33-25

wiiznokes:
@jacobgkau fixed by last commit

mmstick (approved):

jacobgkau (changes requested):
The `test` and `settings` queries are both working as expected now.

Standard regression testing failed. GNOME Settings panels are not showing up in search anymore with this update. This can be tested with e.g. `appearance` for the Appearance panel or `wifi` for the Wi-Fi panel.

wiiznokes:
I suppose this is a popos only thing? Would you be able to provide an example of `.desktop` file that don't work? I'm on Fedora

mmstick:
gnome-control-center desktop entries in the GNOME session should be searchable.

jacobgkau:
> Would you be able to provide an example of `.desktop` file that don't work?

Here's one, for the Appearance page:

```
jacob@serw13:~$ cat /usr/share/applications/gnome-desktop-appearance-panel.desktop
[Desktop Entry]
Name=Appearance
Comment=Set light or dark theme
Exec=gnome-control-center desktop-appearance
# FIXME
# Translators: Do NOT translate or transliterate this text (this is an icon file name)!
Icon=preferences-pop-desktop-appearance
Terminal=false
Type=Application
NoDisplay=true
StartupNotify=true
Categories=GNOME;GTK;Settings;DesktopSettings;X-GNOME-Settings-Panel;X-POP-DesktopSettings;
OnlyShowIn=GNOME;Unity;
# Translators: Search terms to find the Privacy panel. Do NOT translate or localize the semicolons! The list MUST also end with a semicolon!
Keywords=
X-Ubuntu-Gettext-Domain=gnome-control-center-2.0
jacob@serw13:~$
```

wiiznokes:
I see. Last commits should solve this issue. 

The current behavior seems odd to me. The launcher will search in desktop entries that set `NoDisplay=true`, which is technically fine, but will discard any entry that set `NoDisplay=true` and don't have an `OnlyShowIn` field. This is the case for example for applets entries.

I'm not sure if applets should use the desktop enty spec here. It could potentially impact other DEs if i understand correctly. cc @mmstick

mmstick:
The previous logic was based on the session. If you are in the GNOME session, then desktop entries which use `OnlyShowIn=GNOME` will appear regardless of `NoDisplay=true`.

wiiznokes:
Yes, i guess this is fine, even if the launcher could miss some useful entry of normal app (i don't have an example but an app not specific to a DE could use `NoDisplay=true`.

This means another launcher could choose to display this entry, and then, cosmic applets shows up.

mmstick:
Yeah, we similarly want cosmic settings pages to appear in the launcher, but not in the app library.

wiiznokes:
Yes, but my point was related to cosmic applets. AFAIU, cosmic applets are not meant to be presented in launchers or app library

wiiznokes:
And the only things that prevent them to being displayed are `NoDisplay=true` and no `OnlyShowIn` field, which is not in the spec, if i'm not mistaken

mmstick:
It's what GNOME is doing, so we're working with behaviors that are already being relied upon.

wiiznokes:
I see. The last 2 commits implements this behavior

mmstick (approved):

jacobgkau (changes requested):
Found another issue on 6ec06ef during standard regression testing. Searching for `extensions` should only return one entry. Before:

Screenshot from 2024-07-11 15-35-04

With this branch, there are two results for that app:

Screenshot from 2024-07-11 15-35-32

This .desktop file comes from `/usr/share/applications/org.gnome.Extensions.desktop`:

```
[Desktop Entry]
Type=Application
Name=Extensions
# Translators: Do NOT translate or transliterate this text (this is an icon file name)!
Icon=org.gnome.Extensions
Comment=Configure GNOME Shell Extensions
Exec=/usr/bin/gnome-extensions-app --gapplication-service
DBusActivatable=true
Categories=GNOME;GTK;Utility;
OnlyShowIn=GNOME;
X-Ubuntu-Gettext-Domain=gnome-shell
```

wiiznokes:
Damn, this will never end :sob: 

I don't really know why there is 2 `extensions` now. I'm using the same logic to deduplicate entries, e.i, only with the appid. I made a test on my machine and it seems to deduplicate as intented. I think it would help if you can provide the other `org.gnome.Extensions.desktop `

Thanks

jacobgkau:
Aha. I found this issue from just before this item was added to our QA checklist: https://github.com/pop-os/shell/issues/293 

In addition to the `org.gnome.Extensions.desktop` that I posted before, there is also an `org.gnome.Shell.Extensions.desktop` installed to the same location:

```
[Desktop Entry]
Type=Application
# Keep in sync with subprojects/extensions-app
Name=Extensions
# Translators: Do NOT translate or transliterate this text (this is an icon file name)!
Icon=org.gnome.Shell.Extensions
# Never launch this, just provide name+icon to portal dialog
Exec=false
OnlyShowIn=GNOME;
NoDisplay=true
X-Ubuntu-Gettext-Domain=gnome-shell
```

I do want to point out that if existing behaviors are breaking the spec, we could probably consider following the spec if we fix those behaviors in their respective other components (i.e. patch the .desktop files in gnome-shell/gnome-control-center/etc). We just can't ship the update when it would break user-facing behaviors on the checklist.

As for why this duplicate wasn't showing up before, I think it's because the exec line is set to false. If I change the exec line for e.g. LibreOffice Writer's .desktop file to false, it shows with this branch and doesn't show on master. So that's the regression we're looking at now.

wiiznokes:
> As for why this duplicate wasn't showing up before, I think it's because the exec line is set to false.

Yep, that must be it. I don't think this is in the spec but that's not surprising coming from Gnome

wiiznokes:
Should be hopefully good now

mmstick (approved):

jacobgkau (approved):
The `test` and `settings` queries are still working as expected.

<details>
<summary>Standard regression testing passed:</summary>

</details>

PR #221, opened 2024-06-11
Bookmark/history firefox

wiiznokes:
This plugin allow to search through your firefox bookmarks (and history)!

## Feature
- sorted by last visited
- firefox support (maybe someone could add chrome in the future, thus the name)
- history plugin

You can test it with `b: query` and `h: query`. History is limited to 2000 entries. You will need firefox as default browser to use the plugin. If it's not the case, the plugin will fall back to empty bookmarks/history.

todo:
- [x]  fetch icon from internet, or at least find a better default icon
- [ ] (depend on #219)

screenshot-2024-06-11-00-07-27


wiiznokes:
I'm going to close this, because I don't like the current approach of using the private sql database of firefox. In a perfect world, this shouldn't even be possible, because the db should be encrypted or owned by Firefox.

Plasma use another approach that i find cleaner. It integrate using [this extension](https://github.com/KDE/plasma-browser-integration/tree/master).

This should be possible to reuse it in Cosmic (because no one would want to remake this extension for every browser anyway.
> The plasma-browser-integration package must be installed for this extension to work. It will be available from your distribution's package manager.

Maybe this package could be replaced tho. This should also allow to search through open tabs

wiiznokes:
Its possible to query the history using dbus in plasma `dbus-send --session --print-reply --dest=org.kde.plasma.browser_integration /HistoryRunner org.kde.krunner1.Match string:"you"
`

PR #225, opened 2024-07-03
Feat/alt tab cycle

wiiznokes:
based on #219 

impl https://github.com/pop-os/cosmic-launcher/issues/135

With this PR, the service is not intended to be restarted everytime the cosmic-launcher is shown, because the toplevel plugin needs to keep track of some states.

So i just added a field in the plugin config:

```rs
#[serde(default)]
pub long_lived: bool,
```

For plugins that need to always running. 

And this variant Frontend -> Service/Plugin
```rs
/// The frontend was closed and the service should release resources
/// and prepare for the next query
Close,
```

I believe the service was written with this usecase in mind, so there isn't much change.

Also, i had to deactivate the sort when the query is empty.

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -5,6 +5,8 @@ use pop_launcher_toolkit::plugins;
 use pop_launcher_toolkit::service;
 
 use mimalloc::MiMalloc;
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::EnvFilter;
 
 #[global_allocator]
 static GLOBAL: MiMalloc = MiMalloc;
@@ -37,7 +39,10 @@ async fn main() {
     }
 }
 
+// todo: support journald once this issue is resolved: https://github.com/tokio-rs/tracing/issues/2348
 fn init_logging(cmd: &str) {
+    use tracing_subscriber::{fmt, Registry};
+
     let logdir = match dirs::state_dir() {
         Some(dir) => dir.join("pop-launcher/"),
         None => dirs::home_dir()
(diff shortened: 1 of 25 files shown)


bin/src/main.rs
```diff
@@ -49,15 +52,33 @@ fn init_logging(cmd: &str) {
 
     let logfile = std::fs::OpenOptions::new()
         .create(true)
-        .truncate(true)
+        .truncate(false)
```
wash2, line 55: Why set truncate to `true`?

service/src/priority.rs
```diff
@@ -48,7 +48,9 @@ impl PartialEq for Priority {
 impl Eq for Priority {}
 
 impl PartialOrd for Priority {
+    #[allow(clippy::non_canonical_partial_ord_impl)]
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        // todo: what is going on here ?
```
wash2, line 53: Is there something specific that is confusing here about the comparison? If you are asking about the second element in the tuple, it is adjusting the comparison of the match scores when an element is used more recently or more often.

plugins/src/cosmic_toplevel/mod.rs
```diff
@@ -67,20 +70,37 @@ pub async fn main() {
...
                     }
                     ToplevelEvent::Update(handle, info) => {
-                        if let Some(t) = app.toplevels.iter_mut().find(|t| t.0 == handle) {
-                            t.1 = info;
+                        debug!("Update {}", &info.app_id);
+                        debug!("Update {:?}", &info.state);
+
+                        if let Some(pos) = app.toplevels.iter().position(|t| t.0 == handle) {
+                            if info.state.contains(&State::Activated) {
+                                tracing::warn!("Update {:?}: push front", &info.app_id);
```
wash2, line 95: Does this need to be `warn` severity?

service/src/priority.rs
wiiznokes: I just added it because i added `#[allow(clippy::non_canonical_partial_ord_impl)]` to remove the warning without understanding the code. Btw, this code is part of this pr #219

bin/src/main.rs
wiiznokes: Probably to not make a big log file. But this make thing very inconvenient for debugging, because this code is run each time cosmic-launcher is toggled. Settings this to false didn't change anything for me somehow.
But i'm using journald now, but it doesn't support span unfortunately, which doesn't make it a perfect solution.

wash2: Ahh sorry, I meant to say false. I think it should probably remain false, so that the log file doesn't keep growing.

wiiznokes: There is a condition some lines under:
```rs
if let Ok(meta) = file.metadata() {
            if meta.len() > 10000 {
                let _ = file.set_len(0);
            }
        }
```

PR #227, opened 2024-07-13
feat: alt tab cycle

wiiznokes:
Old PR #225 

impl https://github.com/pop-os/cosmic-launcher/issues/135

With this PR, the service is not intended to be restarted everytime the cosmic-launcher is shown, because the toplevel plugin needs to keep track of some states.

So i just added a field in the plugin config:

```rs
#[serde(default)]
pub long_lived: bool,
```

For plugins that need to always running. 

And this variant Frontend -> Service/Plugin
```rs
/// The frontend was closed and the service should release resources
/// and prepare for the next query
Close,
```

I believe the service was written with this usecase in mind, so there isn't much change.

Also, i had to deactivate the sort when the query is empty.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
+checksum = "f5fb1d8e4442bd405fdfd1dacb42792696b0cf9cb15882e5d097b742a676d375"
 dependencies = [
  "gimli",
 ]
 
 [[package]]
-name = "adler"
-version = "1.0.2"
+name = "adler2"
+version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
 name = "aho-corasick"
@@ -43,9 +43,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.86"
+version = "1.0.89"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b3d1d046238990b9cf5bcde22a3fb3584ee5cf65fb2765f454ed428c7a0063da"
+checksum = "86fdf8605db99b54d3cd748a44c6d04df638eb5dafb219b135d0149bd0db01f6"
 
 [[package]]
 name = "async-broadcast"
@@ -73,13 +73,13 @@ dependencies = [
 
 [[package]]
 name = "async-executor"
-version = "1.12.0"
+version = "1.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c8828ec6e544c02b0d6691d21ed9f9218d0384a82542855073c2a3f58304aaf0"
+checksum = "30ca9a001c1e8ba5149f91a74362376cc6bc5b919d92d988668657bd570bdcec"
 dependencies = [
  "async-task",
  "concurrent-queue",
- "fastrand 2.1.0",
+ "fastrand 2.1.1",
  "futures-lite 2.3.0",
  "slab",
 ]
@@ -117,9 +117,9 @@ dependencies = [
 
 [[package]]
 name = "async-io"
-version = "2.3.3"
+version = "2.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0d6baa8f0178795da0e71bc42c9e5d13261aac7ee549853162e66a241ba17964"
+checksum = "444b0228950ee6501b3568d3c93bf1176a1fdbc3b758dcd9475046d30f4dc7e8"
 dependencies = [
  "async-lock 3.4.0",
  "cfg-if",
@@ -127,11 +127,11 @@ dependencies = [
  "futures-io",
  "futures-lite 2.3.0",
  "parking",
- "polling 3.7.1",
- "rustix 0.38.34",
+ "polling 3.7.3",
+ "rustix 0.38.37",
  "slab",
  "tracing",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -175,12 +175,12 @@ dependencies = [
 
 [[package]]
 name = "async-process"
-version = "2.2.3"
+version = "2.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f7eda79bbd84e29c2b308d1dc099d7de8dcc7035e48f4bf5dc4a531a44ff5e2a"
+checksum = "63255f1dc2381611000436537bbedfe83183faa303a5a0edaf191edef06526bb"
 dependencies = [
  "async-channel",
- "async-io 2.3.3",
+ "async-io 2.3.4",
  "async-lock 3.4.0",
  "async-signal",
  "async-task",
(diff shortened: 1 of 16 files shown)


mmstick (approved):

jacobgkau (commented):
This is doing something, but the behavior seems like it needs to be tweaked a little.

1. https://github.com/pop-os/cosmic-launcher/issues/135 specifies pressing/releasing Alt-Tab should switch back and forth between the two most recent windows. I'm seeing that work sporadically, but usually, the currently selected window shows up first in the list, so pressing/releasing Alt-Tab doesn't change the window at all.
2. I'm seeing one window (titled `Oracle VM VirtualBox Manager`) basically stuck at the second or third position even after switching between many other windows (from Firefox and COSMIC Terminal) and not touching that one. So it seems like the recency-based sorting isn't working quite right, or there's somehow a way for another app/window to bypass it.

jacobgkau:
Reading https://github.com/pop-os/cosmic-launcher/issues/135, I do see it specifies applications and not windows. Let me know if we need to get UX clarification on that, if it's supposed to be working with applications right now. I think issue #1 is independent of that, based on the behavior.

wiiznokes:
Just to be sure, did you test with https://github.com/pop-os/cosmic-launcher/pull/150 ? With this PR, alt+tab will select the second app.

I was able to reproduce 2.

jacobgkau:
Ah, no I didn't. I'll get that built on CI for testing.

wiiznokes:
I think it require launcher to be located in ../launcher currently. I will change that once i fix point 2

wiiznokes:
I think there is a bug in cosmic-comp that make this PR not working:

code for the log:
```rs
ToplevelEvent::Update(handle, info) => {
                        warn!("Update {}: {:?}", &info.app_id, &info.state);
```

Output when selecting konsole -> cosmic-edit -> virtualbox -> firefox (using cosmic-launcher)

```
juil. 19 01:22:31 fedasus pop-launcher[22633]: Update org.kde.konsole: {}
juil. 19 01:22:31 fedasus pop-launcher[22633]: Update com.system76.CosmicEdit: {Activated}

juil. 19 01:22:41 fedasus pop-launcher[22633]: Update com.system76.CosmicEdit: {}
juil. 19 01:22:41 fedasus pop-launcher[22633]: Update VirtualBox Manager: {Activated}
juil. 19 01:22:41 fedasus pop-launcher[22633]: Update com.system76.CosmicEdit: {Activated}
juil. 19 01:22:41 fedasus pop-launcher[22633]: Update VirtualBox Manager: {}
juil. 19 01:22:41 fedasus pop-launcher[22633]: Update com.system76.CosmicEdit: {}
juil. 19 01:22:41 fedasus pop-launcher[22633]: Update VirtualBox Manager: {Activated}

juil. 19 01:22:45 fedasus pop-launcher[22633]: Update VirtualBox Manager: {}
juil. 19 01:22:45 fedasus pop-launcher[22633]: Update org.mozilla.firefox: {Activated}
juil. 19 01:22:45 fedasus pop-launcher[22633]: Update com.system76.CosmicEdit: {Activated}
juil. 19 01:22:45 fedasus pop-launcher[22633]: Update org.mozilla.firefox: {}
juil. 19 01:22:45 fedasus pop-launcher[22633]: Update com.system76.CosmicEdit: {}
juil. 19 01:22:45 fedasus pop-launcher[22633]: Update org.mozilla.firefox: {Activated}
```

I'm not sure this protocol is intended for this use case tho

cc @Drakulix

jacobgkau (changes requested):
This seems to be working pretty well now.

The recency is not taking into account window activations that are a result of switching workspaces. This causes a couple of issues:

In this first video, the OBS window is never suggested as the second-most-recent window, even when I go from it directly to another window (by switching workspaces).

https://github.com/user-attachments/assets/a3ef5bef-c486-4f93-9944-32bb62472569

In this second video, even when I select a second window on that second workspace, alt-tab brings me back to the first workspace, instead of back to the previously-selected OBS window.

https://github.com/user-attachments/assets/9f240e0c-80b8-4563-a500-58621269c0b6

I think the only thing that needs to change is for the recency to take into account window activations caused by workspace switching. (Alternatively, switching could be isolated per-workspace to avoid this issue, but I don't think that was the intended design.)

wiiznokes:
probably a cosmic-comp issue @Drakulix

Drakulix:
> probably a cosmic-comp issue @Drakulix

No it isn't. The current workspace doesn't affect the `Activated` state. You need to use the `cosmic-workspace` protocol as well and track which workspace an application is on to fix this.

wiiznokes:
How would i know which top_level is active just by knowing witch workspace was activated ? Let's says there are multiple windows in workspace B. And also the case where all windows in workspace B are minimized, i guess we don't want to bump any top_level in the list.

Drakulix:
> How would i know which top_level is active just by knowing witch workspace was activated ? Let's says there are multiple windows in workspace B. And also the case where all windows in workspace B are minimized, i guess we don't want to bump any top_level in the list.

The one with the `Activated` state on said workspace of course. (Which isn't necessarily the last window to get `Activated`, just the last on said workspace, that was `Activated`.)

wiiznokes:
> The one with the Activated state on said workspace of course. (Which isn't necessarily the last window to get Activated, just the last on said workspace, that was Activated.)

I try to implement it in this branch https://github.com/wiiznokes/launcher/tree/alt-tab-with-workspace-support-1.
I does not work and i'm not sure why. I think the current behavior of both protocols make this impossible, tho, i might miss understand smtg

Drakulix:
> I try to implement it in this branch https://github.com/wiiznokes/launcher/tree/alt-tab-with-workspace-support-1. I does not work and i'm not sure why. I think the current behavior of both protocols make this impossible, tho, i might miss understand smtg

I am sorry, but I am not going to debug your branch and you are not providing me with a bunch of information.
What do you mean with "it does not work". Did you run with `WAYLAND_DEBUG=1` to figure out what events you get and when? Do you have any specific questions about the implementation.

wiiznokes:
Ok, sorry about that. I figured out the bug in my code that was causing no sense (using `position` on a reversed iterator :facepalm:). I work better now, except in this case:

1. App A in workspace X is active
2. Focus an app B in workspace Y
3. Focus an app C in workspace X

result: The list will be C -> A -> B instead of C -> B -> A, because both `WorkspaceHandler::done` and `ToplevelInfoHandler::update_toplevel` will be called

Drakulix:
> Edit: And `WorkspaceHandler::done` is called first. If it was guaranteed to be called after`ToplevelInfoHandler::update_toplevel`, that could solve this

From a cosmic-comp perspective we first swap workspaces and fixup focus afterwards, which... yeah.. I get how this is a hassle.

From a protocol perspective it is quite difficult to guarantee any specific order between the two protocols.

It might be easier to remove the `Activated`-state from the original window on workspace swap, so you can *just* rely on `cosmic-toplevel` without needing `cosmic-workspaces`.

wiiznokes:
> It might be easier to remove the Activated-state from the original window on workspace swap, so you can just rely on cosmic-toplevel without needing cosmic-workspaces.

But in my example, if the app A get focused instead of C, cosmic-toplevel will not be trigger.

wiiznokes:
@Drakulix Just to be sure, do you agree that cosmic-comp need to re send the `Activated` state when switching workspace, even we it has not _changed_ ?

Drakulix:
> @Drakulix Just to be sure, do you agree that cosmic-comp need to re send the `Activated` state when switching workspace, even we it has not _changed_ ?

Yes. Though this would also mean we remove the `Activated` state on workspace switch from the previous workspace. But that shouldn't be an issue, right?

wiiznokes:
> Yes. Though this would also mean we remove the Activated state on workspace switch from the previous workspace. But that shouldn't be an issue, right?

I don't think so. I think is it will better overall this way

Drakulix:
> > Yes. Though this would also mean we remove the Activated state on workspace switch from the previous workspace. But that shouldn't be an issue, right?
> 
> I don't think so. I think is it will better overall this way

Great. I think this *should* fix it: https://github.com/pop-os/cosmic-comp/pull/889

simonvanderveldt:
Is my understanding correct that this has been carried/fixed via https://github.com/pop-os/cosmic-launcher/pull/225? There's no link to https://github.com/pop-os/cosmic-launcher/issues/135 nor to this PR in #225 , so I'm not really sure.

simonvanderveldt:
> Yeah your understanding is correct

OK, thanks for the confirmation and for the implementation of course :+1:

PR #229, opened 2024-07-18
fix(toplevel): require higher toplevel match score

### LINKED ISSUE #228: [COSMIC Epoch; regression] Search results don't show apps when too many windows are open
This seems to be caused by 172532056b404923b0bc068b524b23e81a371f18. It does not happen on ab7a71a057e3e2ce5fc200b056799a07cf3a9bb5. (FYI, @wiiznokes in case you have an idea of why this might be happening.)

Previously, search results would return application names even if a lot of windows are open:

screenshot-2024-07-17-14-20-20

On the latest commit, searching for the same string, I'm getting a bunch of irrelevant (but presumably closest-matching) window results, and no application results:

screenshot-2024-07-17-14-21-46

This essentially breaks launching apps with the launcher once more than 8 windows are open.

The good news is, this is not happening in the GNOME session (which is why I didn't catch it during regression testing).

wash2:
Raising the required match score fixes #228 based on my testing and matching individual words seems to better match window titles.

### DIFF
--- a/plugins/src/cosmic_toplevel/mod.rs
+++ b/plugins/src/cosmic_toplevel/mod.rs
@@ -18,6 +18,7 @@ use pop_launcher::{
     Request,
 };
 use std::borrow::Cow;
+use std::iter;
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 
 use self::toplevel_handler::{toplevel_handler, ToplevelAction, ToplevelEvent};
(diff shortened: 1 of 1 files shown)


mmstick (approved):

jacobgkau (approved):
Confirmed this fixes https://github.com/pop-os/launcher/issues/228, or at least improves it significantly. Searching `ter` still gives me all the Firefox windows, but `term` now returns the COSMIC Terminal app launcher, followed by other apps.

Looking at the code, this shouldn't touch the GNOME session at all, but I did regression testing to be safe, and it passed. 

<details>
<summary>Checklist:</summary>

</details>

Issue #230, opened 2024-07-31
reproducible builds issue


bmwiedemann:
While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds) (sponsored by the NLnet NGI0 fund), I found that our `pop-launcher` package build produces different results between a 1-core-VM and a VM with 2+ cores. This could be an indicator of some race-condition or poorly defined build-dependencies.

https://rb.zq1.de/other/pop-launcher/strings.diff.txt
shows that some symbols have variations in the ID at the end.
e.g.
```diff
-_ZN4core3ptr43drop_in_place$LT$zvariant..value..Value$GT$17he5bf3c79d5d19a4aE.28032
+_ZN4core3ptr43drop_in_place$LT$zvariant..value..Value$GT$17he5bf3c79d5d19a4aE.28050
```

The two full binaries are also available in the same directory.

mmstick:
Rust projects will always have different binaries when building with a different number of [codegen units](https://doc.rust-lang.org/nightly/rustc/codegen-options/index.html#codegen-units). You can configure your build system to 1 if you want it to use only one.

bmwiedemann:
Indeed. #231 seems to avoid that issue.

PR #231, opened 2024-07-31
Use a single codegen-unit for compilation

bmwiedemann:
Use a single codegen-unit for compilation
as suggested in
https://nnethercote.github.io/perf-book/build-configuration.html#codegen-units

to achieve reproducible builds.
It can also enhance optimizations of the produced binary.

This patch was done while working on reproducible builds for openSUSE, sponsored by the NLnet NGI0 fund.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ serde_json.workspace = true
 serde_with.workspace = true
 
 [profile.release]
+codegen-units = 1
 lto = "fat"
 panic = "abort"
 


jacobgkau:
@mmstick From your comment in https://github.com/pop-os/launcher/issues/230, it sounded like you were suggesting e.g. openSUSE could configure their build environment to use a set number of codegen units. Does it make sense to do this here at the upstream project level?

mmstick:
It's something that OpenSUSE should do in their build environment if they want this. It'll slow down compile times for everyone else.

PR #234, opened 2024-08-05
Use more generic shell script shebang

### LINKED ISSUE #233: `sh` shebang in bundled scripts
The bundled scripts (for logging out, rebooting, etc.) use `#!/usr/bin/sh` as a shebang which doesn't work on NixOS (or other distributions that are not [FHS](https://de.wikipedia.org/wiki/Filesystem_Hierarchy_Standard)-compliant). Replacing it by `#!/usr/bin/env sh` should work everywhere. If that's fine with you, I'll create a PR to update the scripts. If not, just close the issue and we can patch this downstream.

Triton171:
Fixes #233.

### DIFF
--- a/scripts/session/session-bios.sh
+++ b/scripts/session/session-bios.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/sh
+#!/usr/bin/env sh
 #
 # name: Enter BIOS
 # icon: system-restart
--- a/scripts/session/session-lock.sh
+++ b/scripts/session/session-lock.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/sh
+#!/usr/bin/env sh
 #
 # name: Lock
 # icon: locked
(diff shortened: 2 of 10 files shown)


mmstick (approved):

jacobgkau (approved):
Researching the implications of using `env` instead of directly using `sh`, it seems like a fairly niche case that would benefit, but I don't see any concrete reasons not to. (My understanding is that this will use the first `sh` in the `$PATH`. This could have security implications, but could just as easily be something intentionally configured.)

Running shell commands and shell scripts with the `:` prefix still works.

On Pop!_OS 22.04 and 24.04, I did find that the `sh` command and the `/usr/bin/sh` command (even in a terminal) started a shell that reported `/bin/bash` (when checking with e.g. `echo $SHELL`); on those systems, `sh` is symlinked to `dash` by default. This launcher change does not alter the behavior in the launcher.

<details>

</details>

jacobgkau:
This might have to wait a little bit before merging, since we're working on tagging COSMIC Epoch Alpha 1 this morning. (This may not make Alpha 1, but if not, it should be merged after that's finished.)

PR #239, opened 2024-09-10
fix(desktop_entries): Respect `NoDisplay` override precedence

joshuamegnauth54:
Closes: pop-os/cosmic-epoch#905

User defined desktop entries should override the system's applications. The old logic continued to parse apps which overrode user preference, such as with the `NoDisplay` key.

## Updates (force pushes)
* ~~Removed old `NoDisplay` code~~
* Revert the above
* Small clean up

### DIFF
--- a/plugins/src/desktop_entries/mod.rs
+++ b/plugins/src/desktop_entries/mod.rs
@@ -76,9 +76,15 @@ impl<W: AsyncWrite + Unpin> App<W> {
         let desktop_entries = DesktopEntry::from_paths(paths, &locales)
             .filter_map(|de| {
                 de.ok().and_then(|de| {
-                    if deduplicator.contains(de.appid.as_ref()) {
+                    // Treat Flatpak and system apps differently in the cache so they don't
+                    // override each other
+                    let appid = de.flatpak().unwrap_or_else(|| de.appid.as_ref());
+                    if deduplicator.contains(appid) {
                         return None;
                     }
+                    // Always cache already visited entries to allow overriding entries e.g. by
+                    // placing a modified copy in ~/.local/share/applications/
+                    deduplicator.insert(appid.to_owned());
 
                     if de.name(&self.locales).is_none() {
                         return None;
@@ -128,12 +134,14 @@ impl<W: AsyncWrite + Unpin> App<W> {
                                 return None;
                             }
                         }
-                    } else {
-                        if de.no_display() {
-                            return None;
-                        }
                     }
-                    deduplicator.insert(de.appid.to_string());
+                    // Treat `OnlyShowIn` as an override otherwise do not show if `NoDisplay` is true
+                    // Some desktop environments set `OnlyShowIn` and `NoDisplay = true` to
+                    // indicate special entries
+                    else if de.no_display() {
+                        return None;
+                    }
+
                     Some(de)
                 })
             })


wiiznokes:
Have you tested your changes? I can see a few problem with this

1.
```rust
if let Some(flatpak) = de.flatpak() {
                        flatpak
                    } else {
                        de.appid.as_ref()
                    }
```
will return the same string for my app (and probably all ?)
```
cat /var/lib/flatpak/exports/share/applications/io.github.wiiznokes.fan-control.desktop 
[Desktop Entry]
...
X-Flatpak=io.github.wiiznokes.fan-control
```
2. No display was already checked 
3. `deduplicator.insert` should be at the end, when we are sure we will use this entry

wiiznokes:
btw, the issue you linked reference this repo: https://github.com/pop-os/cosmic-applibrary

joshuamegnauth54:
Thank you for your comments! They're useful because I can clarify the fix. :smile_cat: 

Yep! I tested it. Originally I only added a filter for `NoDisplay` but that proved to be insufficient hence the filtering changes.

1. Yes, for some apps that block of code would return the same string. That's okay and intentional. Some desktop entries (particularly Flatpaks) have different names for `X-Flatpak` and `Name`, so this would preserve both entries while allowing users to override as needed.

2. The condition I added and the code you're referring to are different. The original `NoDisplay` check only triggers if `OnlyShowIs` is _not_ defined, but it's possible that both `OnlyShowIn` and `NoDisplay` are both defined which would override the original `NoDisplay` check. `NoDisplay` should override `OnlyShowIn` if the user doesn't want to show the desktop entry. I left the original check in because I didn't want to change that behavior in case there was a reason for it lol.

3. As per the code comments, I changed that behavior because it wasn't entirely correct. Users may locally define desktop entries. The previous behavior would yield the system's entries without overriding, so filtering should be done. To test this, copy a desktop entry into the local applications folder and set `NoDisplay`. The entry would still show because the loop would continue onto the system entries.

As for applibrary, it seemed prudent to fix the issue in this repo so it would propagate to other services that may use the launcher. The launcher filters already, so I thought it would better to improve said filtering here instead of on a per app basis.

wiiznokes:
> Yes, for some apps that block of code would return the same string. That's okay and intentional. Some desktop entries (particularly Flatpaks) have different names for X-Flatpak and Name, so this would preserve both entries while allowing users to override as needed.

You know probably know better what `X-Flatpak` is assigned to, but i was refering to the appid (e.i., the name of the deskop entry without the `.desktop`)

>  I left the original check in because I didn't want to change that behavior in case there was a reason for it lol.

I think this change is fine, but you should remove the previous check, because it could cause confusion, and it have a performance penalty for nothing

> As per the code comments, I changed that behavior because it wasn't entirely correct. Users may locally define desktop entries. The previous behavior would yield the system's entries without overriding, so filtering should be done. To test this, copy a desktop entry into the local applications folder and set NoDisplay. The entry would still show because the loop would continue onto the system entries.

That make total sense, thanks for clarifying 

> As for applibrary, it seemed prudent to fix the issue in this repo so it would propagate to other services that may use the launcher. The launcher filters already, so I thought it would better to improve said filtering here instead of on a per app basis.

Make sense, tho i'm not aware of anything using this launcher expect the cosmic-launcher.

jacobgkau:
> i'm not aware of anything using this launcher expect the cosmic-launcher.

The third-party [onagre-launcher](https://github.com/onagre-launcher/onagre) uses pop-launcher, in addition to the pop-shell and cosmic-launcher frontends.

git-f0x:
Wouldn't the launcher respecting `NoDisplay` hide the COSMIC Settings page shortcuts (e.g. when typing "Panel")?

jacobgkau (changes requested):
@git-f0x is correct. Currently, this change results in COSMIC Settings shortcuts no longer showing up in COSMIC Epoch, and also results in GNOME Settings panels not showing up in the 22.04 LTS GNOME session.

In order for this to move forward, another solution would need to be introduced to show the settings panels. I wonder if an `OnlyShowIn` specifically for the launcher would be more appropriate than using `NoDisplay` in this way? Or a plugin for the launcher specifically to show settings pages, rather than having them be `.desktop` files. (Or, as @git-f0x suggested, another way of hiding entries from the launcher could be introduced to at least address the usecase.)

joshuamegnauth54:
@git-f0x That's a good catch! I didn't notice that at all.

@jacobgkau I researched this issue a bit more, and it turns out it's messier than I expected! Other app launchers (I tried or checked the source code for [rofi](https://github.com/davatorium/rofi), [yofi](https://github.com/l4l/yofi), [cerebro](https://github.com/cerebroapp/cerebro), [albert](https://github.com/albertlauncher/), and then I felt too lazy to continue).

`yofi` [skips `NoDisplay`](https://github.com/l4l/yofi/blob/09901e75cbdf2147553ab888adde480e57baa0d1/src/desktop.rs#L126-L129) entries:

```rust
        if main_section.attr("NoDisplay") == Some("true") {
            log::trace!("Skipping NoDisplay entry {:?}", dir_entry);
            return None;
        }
```

`rofi`/`wofi` also skips `NoDisplay` entries. I checked this myself instead of reading the source.

`cerebro` [skips](https://github.com/cerebroapp/plugins/blob/main/cerebro-basic-apps/src/linux.js#L117-L135) `NoDisplay` entries:
```javascript
export const formatPath = (filePath) => {
  const parsedData = parseDesktopFile(filePath, {
    name: 'Name',
    description: 'Comment',
    exec: 'Exec',
    hidden: 'NoDisplay',
    icon: 'Icon'
  })
  const filename = path.basename(filePath)
  return {
    ...parsedData,
    filename,
    icon: findIcon(parsedData.icon),
    hidden: !!parsedData.hidden || !parsedData.exec,
    id: getId(filePath),
    name: parsedData.name || filename.replace(/\.(desktop)/, ''),
    path: filePath
  }
}
```

And then in [getAppsList()](https://github.com/cerebroapp/plugins/blob/main/cerebro-basic-apps/src/initializeAsync.js#L25-L27):
```javascript
  return Promise.all(promises).then(apps => (
    uniq(flatten(apps)).map(formatPath).filter(app => !app.hidden)
  ))
```

Albert also [skips it](https://github.com/albertlauncher/plugins/blob/f9a7f8ba79ebdf9742781df8241213064db883e7/applications/src/xdg/application.cpp#L24-L28):
```c++
    // NoDisplay - boolean, must not be true
    try {
        if (p.getBoolean(root_section, QStringLiteral("NoDisplay")))
            throw runtime_error("Desktop entry excluded by 'NoDisplay'.");
    } catch (const out_of_range &) { }
```

Here is a non-exhaustive list of entries that define `NoDisplay=true`:
xdg-desktop-portal-gtk.desktop
java-java-openjdk.desktop
org.openusd.usdview.desktop
nsxiv.desktop
qemu.desktop
org.gnome.Zenity.desktop
geoclue-where-am-i.desktop
org.pwmt.zathura-pdf-poppler.desktop
code-oss-url-handler.desktop
org.freedesktop.Xwayland.desktop
gcr-prompter.desktop
wine.desktop
org.gnupg.pinentry-qt5.desktop
gcr-viewer.desktop
com.system76.CosmicSettings.Users.desktop
com.system76.CosmicBackground.desktop
com.system76.CosmicSettings.Desktop.desktop
com.system76.CosmicAppletNotifications.desktop
com.system76.CosmicAppletTiling.desktop
com.system76.CosmicSettings.Time.desktop
gnome-users-panel.desktop
gnome-network-panel.desktop
org.gnome.Shell.Extensions.desktop
com.system76.CosmicNotifications.desktop
xdg-desktop-portal-gtk.desktop
flatpak-installer.desktop
org.gnome.Shell.desktop

joshuamegnauth54:
I'll try to figure out a better solution in the meantime. :cat:

git-f0x:
>Or, as @git-f0x suggested, another way of hiding entries from the launcher could be introduced to at least address the usecase.

I was more referring to the `.desktop` files for the Settings pages to potentially drop `NoDisplay` (since they already have `OnlyShowIn=COSMIC`), and instead use a custom key (maybe a better name would be `X-OnlyShowInLauncher`) that is treated as `NoDisplay` by COSMIC components other than the Launcher.
This would enable hiding `.desktop` entries by users in a standard way, while keeping current functionality.

joshuamegnauth54:
I reverted the old `NoDisplay` behavior but kept the new precedence code. I also wrote a comment on the reverted `NoDisplay` to explain it for future contributors. It seems like GNOME as well as COSMIC define both `NoDisplay` and `OnlyShowIn` for special cases.

I haven't tested this yet, and unfortunately I have a busy weekend so I'll have to try it on Monday.

joshuamegnauth54:
Okay the fix from a few days ago seems to work. I'm able to launch directly into settings pages again _and_ overrides work as well.

Entries that define both `NoDisplay=true` and `OnlyShowIn=[some DE]` will still show and can't be hidden (yet). In the unlikely case that a user wishes to hide these special entries, they would need to define `Hidden=true`. `launcher` doesn't check for that key yet, but I'll open a PR for that later because I opened a related patch ([pop-os/freedesktop-desktop-entry#35]) that adds a convenience function for that key.

jacobgkau (approved):
This is now working without causing any regressions to expected behavior on Pop!_OS 24.04 with COSMIC.

<details>
<summary>cosmic-launcher regression testing:</summary>

### Launcher

- [X] Search works for COSMIC Settings panels
    - Selecting a settings panel when Settings is open on another workspace does not switch workspaces; not a regression.
- [X] Open windows are sorted above applications (e.g. "web browser")
</details>

Regression testing also passed on Pop!_OS 22.04 with GNOME, including settings panel search.

<details>
<summary>pop-shell regression testing:</summary>

</details>

If I copy a system-wide `.desktop` file from `/usr/share/applications/` to `~/.local/share/applications/` and then comment out the `OnlyShowIn` line within the local copy (leaving the `NoDisplay` line), that icon is now hidden from the launcher search results, where it was not hidden before.

mmstick (approved):

joshuamegnauth54:
@mriganksagar It should be "NoDisplay" not "noDisplay". See: https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html

If that doesn't work, you can open a new issue and I'll take a look at it.

PR #241, opened 2024-09-28
fix: total ordering for Priority

wash2:
this should fix https://github.com/pop-os/cosmic-launcher/issues/191
The weights could be adjusted for recent or frequency score if necessary.


mmstick (approved):

jacobgkau (approved):
This seems to sort results comparably to the current stable version of the launcher. I didn't find any differences. The frequency score is still very weak compared to the recency score.

<details>

### Launcher

</details>

I was not able to recreate https://github.com/pop-os/cosmic-launcher/issues/191 in a Fedora 40 VM with the COPR repo mentioned there, so can't confirm whether this will fix the issue building for distros like that that have Rust 1.81+.

PR #242, opened 2024-09-29
fix: use /usr/libexec/.. for plugins instead of /usr/lib/..

ryanabx:
This commit changes plugins to use /usr/libexec instead of /usr/lib. Since the plugins contain executables, the more apt directory for this need is /usr/libexec. Fedora is requiring this in order to package pop-launcher and cosmic-launcher, so this patch will be used for downstream Fedora, but if upstream would like to adopt this change, I've brought it here just in case! :)


mmstick (dismissed):

mmstick:
Shouldn't be necessary since it communicates to the launcher via IPC over stdin/stdout.

ryanabx:
> Shouldn't be necessary since it (cosmic-launcher) communicates to the pop-launcher sub-process via IPC over stdin/stdout.

Oh I didn't know, thanks!

jacobgkau (approved):
This moves the system-wide `pop-launcher/{plugins,scripts}` directories from `/usr/lib/` to `/usr/libexec/`. There's no change in behavior.

<details>

### Launcher

</details>

jackpot51 (changes requested):
I believe both Arch and NixOS have issues with /usr/libexec. The best thing to do is make it configurable.

ryanabx:
@jackpot51 how's this?

Conan-Kudo:
`/usr/libexec` should be the default since that is the recommended place for the majority of distributions (Fedora, openSUSE, Debian/Ubuntu, etc.) and distributions who want an alternative location can set it otherwise.

wiiznokes:
I wonder how the `POP_LAUNCHER_LIB_DIR` even works since it is not exported in a `build.rs` file.

Except that, i have some suggestion/question
- why all the paths have been duplicated in the code ?
- could we change `option_env!` to `env!`, and declare the `POP_LAUNCHER_LIB_DIR` variable in the justfile. This will make things a bit less magic, and easier for packaging and development i think, because we will just need to call change `lib-path` in the Makefile.

PR #246, opened 2024-10-18
chore: update dependencies and release 1.2.4

mmstick:
Updates all dependencies and creates a 1.2.4 release

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


wash2 (approved):

leviport (approved):
No regressions detected.

---
- [x] All windows on all workspaces appear on launch
- [x] Choosing an app on another workspace moves workspaces and focus to that app
- [x] Launching an application works
- [x] Typing text and then removing it will re-show those windows
- [x] Search works for applications and windows
- [x] Search works for GNOME settings panels
- [x] Search for "Extensions". There should be only one entry.
- [x] The overlay hint correctly highlights the selected window
- [x] Open windows are sorted above applications (e.g. "web browser")
- [x] t: executes a command in a terminal
- [x] : executes a command in sh
- [x] = calculates an equation
- [x] Search results are as expected:

PR #249, opened 2024-12-05
Alt tab with workspace & output support 

wash2:


wash2:
Ok this should be working now with https://github.com/pop-os/cosmic-launcher/pull/225 and the master branch of cosmic-comp. alt+tab reflects focus changes properly now and i don't believe there are regressions.

jacobgkau (dismissed):
I'm still seeing unexpected behavior with this and https://github.com/pop-os/cosmic-launcher/pull/225. See the following:

https://github.com/user-attachments/assets/57dead5a-fe7e-4a7d-880d-ae7d8e0074e1

Steps to recreate the issue:

1. Open two windows on a workspace. Alt-tab switches between them as expected.
2. Switch to a new workspace and open a third window. Alt-tab switches back to the first workspace.
3. Now alt-tab is stuck on a single window.
    - I'm assuming it's supposed to switch us back to the previously activated window on the other workspace, but it should at least switch to the other window on the active workspace instead of doing nothing.

wash2:
> I'm still seeing unexpected behavior with this and [pop-os/cosmic-launcher#225](https://github.com/pop-os/cosmic-launcher/pull/225). See the following:
> 2024-12-06.18-25-10.mp4
> 
> Steps to recreate the issue:
> 
>     1. Open two windows on a workspace. Alt-tab switches between them as expected.
> 
>     2. Switch to a new workspace and open a third window. Alt-tab switches back to the first workspace.
> 
>     3. Now alt-tab is stuck on a single window.
>        
>        * I'm assuming it's supposed to switch us back to the previously activated window on the other workspace, but it should at least switch to the other window on the active workspace instead of doing nothing.

Just double checking, which version of cosmic-comp was running when you tried this? When I try to copy what is done in your video with [7829e76](https://github.com/pop-os/cosmic-comp/commit/7829e7696d276c0b51ea4834986a0a37e40b9ee4) it works as expected.

jacobgkau:
Ah, I was still on e3b41c5 from release, sorry. After adding master to this machine, it is working as expected.

jacobgkau (approved):
Alt-tab is now working as expected on single display/workspace, multiple workspaces, multiple displays, and multiple workspaces/displays. It is currently limited to the last 8 windows. Tested in COSMIC on 24.04 and 22.04.

<details>
<summary>Launcher regression testing passed (22.04 GNOME):</summary>

### Launcher

</details>

plugins/src/desktop_entries/utils.rs
```diff
@@ -36,3 +36,13 @@ pub fn get_description<'a>(de: &'a DesktopEntry<'a>, locales: &[String]) -> Stri
...
+
+// todo: cache
+#[must_use]
+pub fn is_session_cosmic() -> bool {
+    if let Ok(var) = std::env::var("XDG_CURRENT_DESKTOP") {
+        return var.contains("COSMIC");
+    }
+
+    false
+}
```
Drakulix, line 48: I feel like in the future it would be nicer if the plugin.ron files would have an "AlwaysShowIn"-key similar to desktop files, so that other desktops could more easily extend this, if they support the necessary protocols. (Why do the plugins not use desktop-files in the first place?)

Anyway, this is good enough for now.

Drakulix (approved):
LGTM!

PR #254, opened 2025-01-16
refactor: invoke cosmic-osd in session scripts when available

wash2:
This should display a popup dialog with a timer for logout, reboot, and shutdown.


mmstick (approved):

jacobgkau (commented):
This should have gotten a QA review before being merged. The logout, reboot, and shutdown session scripts do appear to be working normally in both GNOME on 22.04 and COSMIC on 24.04.

PR #263, opened 2025-03-08
Allow desktop entries with shell URL handlers to correctly display in launcher

ellieplayswow:
specific example: vscode from flathub

**changes tl;dr**
- the `desktop_entries` plugin adds each appid to a dedupe check really early in the filter function
- some programs may include multiple `.desktop` files, for instance vscode including a `com.visualstudio.code.desktop` and `com.visualstudio.code-url-handler.desktop`
- if one entry otherwise fails the filter (eg: `de.no_display()` set for a URL handler) then any other entries with the same appid get ignored
- this changes that behaviour so we only ignore the appid when we've added it to our list of entries

### DIFF
--- a/plugins/src/desktop_entries/mod.rs
+++ b/plugins/src/desktop_entries/mod.rs
@@ -78,9 +78,6 @@ impl<W: AsyncWrite + Unpin> App<W> {
                     if deduplicator.contains(appid) {
                         return None;
                     }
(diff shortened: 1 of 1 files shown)


ellieplayswow:
fwiw: on my pc, this _only_ adds an extra entry specifically for vscode, everything else seems identical behaviour-wise

wash2 (approved):

PR #268, opened 2025-04-02
chore: latest release of fde for plugins

hepp3n:
Hi,

This PR will update freedesktop-desktop-entry crate to latest released version which is 0.7.9 for now.
I hope it won't break anything.  ALT + TAB works OK, also searching works the same as before (?).


git-f0x:
I think this can be closed now, since it was updated since.

hepp3n:
Yeah I think so :)

PR #269, opened 2025-04-20
desktop_entries: Reload .desktop on search

ljrk0:
Since pop-launcher has been demonized, the desktop_entries plugin isn't relaunched on every search anymore. In consequence, it is not providing up-to-date results when .desktop files have changed (e.g., on un-/installation of a new application) anymore.

Move the app.reload() call into the request loop to trigger a re-index. In a release-build this has been fine performance wise on an 8th Gen i5 laptop device.

Alternatives could include remembering the last time the index was built (initialised to UNIX_EPOCH before the loop) and only conditionally triggering a re-index after a hot-period of 10s since the last run. In our testing this didn't result in any noticeable difference.

NOTE: This has been tested against commit fca3b25552deca3bd64de8bfd1ebb38c45e4fb39 but builds against master too. However, master error'ed with unrelated issues regarding the parsing of CLI arguments.


mmstick (commented):
It'd be better to reload only on an empty search query

ljrk0:
Implemented the requested change + rebased ontop of `master`, but I cannot reproduce the issue since #270 got merged(?). At least not on fedora with ryan's copr pop-launcher package (nightly).

The new behavior seems that the `desktop_entries` plugin doesn't reload the `.desktop` entries anymore, so this behavior is still the same as before:

```
$ /usr/libexec/pop-launcher/plugins/desktop_entries/desktop-entries
# now install vlc
{ "Search": "vlc" }
# doesn't find vlc
```

But it seems that the `pop-launcher` now does some magic to fix this when actually running the launcher itself?

wiiznokes:
You could also add a watcher on the .desktop dirs. Idk if its better performance wise

mmstick (approved):

PR #270, opened 2025-04-30
fix: update freedesktop-desktop-entry

mmstick:
Updates the freedesktop-desktop-entry crate, which should fix some applications not appearing in the search, or not having the correct icon. However, it does alter queries so there may be some differences in search results.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,17 +93,17 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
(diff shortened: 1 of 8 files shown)


jacobgkau (approved):
<details>
<summary>Regression testing passed (22.04 GNOME), including some key sorting sanity checks:</summary>

### Launcher

</details>

Not seeing any issues in a 24.04 Epoch session.

PR #279, opened 2025-09-29
fix: introduce decay factor for keyword weight based on index, and shrink the description match weight.

wash2:
This should fix https://github.com/pop-os/cosmic-launcher/issues/303. We could tweak the values to be smaller if this seems to affect anything else in a negative way.

### DIFF
--- a/service/src/lib.rs
+++ b/service/src/lib.rs
@@ -567,45 +567,6 @@ impl<O: futures::Sink<Response> + Unpin> Service<O> {
         } else {
             active_search.sort_by(|a, b| {
                 // Weight is calculated between 0.0 and 1.0, with higher values being most similar
-                fn calculate_weight(meta: &PluginSearchResult, query: &str) -> f64 {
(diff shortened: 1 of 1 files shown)


mmstick (approved):

jacobgkau (approved):
Testing on 22.04, this fixes https://github.com/pop-os/cosmic-launcher/issues/303.

<details>
<summary>22.04 regression testing passed:</summary>

### Launcher

</details>

I didn't notice any other sorting issues while testing. Of the few questionable orders I saw, they were the same on the currently released version (0e8aa22), and were either caused by my recency/hits or otherwise unrelated to the sorting change being done here.

Also confirmed this fixes https://github.com/pop-os/cosmic-launcher/issues/303 on 24.04.

<details>
<summary>24.04 testing also passed:</summary>

### Launcher

- [X] Search works for COSMIC settings panels
    - `cal` returns Calendar first
    - `shop` returns the COSMIC Store first
</details>

PR #282, opened 2025-12-30
feat(desktop_entries): resolve icons to absolute paths if in standard locations

ZeusCraft10:
Improves icon resolution by checking standard fallback directories like pixmaps and local user icons, handling ~ expansion and missing extensions. This fixes cases where icons for apps like Battle.net and wago.io were missing in frontends that rely on pop-launcher for icon information.

References pop-os/cosmic-epoch#2847


mmstick:
These paths are already handled by the cosmic-freedesktop-icons crate at pop-os/freedesktop-icons when creating an icon handle in libcosmic. If there is a gap in the logic, fixes should go there instead.

